PR #21, opened 2020-10-30
For redhat family make sure to use 'yum list installed' instead of 'yum

roopakparikh:
list'

For redhat command 'yum list' shows list of RPMs by name that can be
fetched from the repos (not just installed). We should be using 'yum
list installed'


abhimanyu-babbar (approved):
👍

PR #24, opened 2020-11-10
Added a flag to ignore the self signed certificate

roopakparikh:
Added a change to ignore certificate checks:
* While downloading the hostagent
* Making keystone and other http calls

Testing: made sure prep-node works

### DIFF
--- a/pkg/pmk/clients.go
+++ b/pkg/pmk/clients.go
@@ -6,6 +6,8 @@ import (
 	"github.com/platform9/pf9ctl/pkg/keystone"
 	"github.com/platform9/pf9ctl/pkg/resmgr"
 	"github.com/platform9/pf9ctl/pkg/cmdexec"
+	"net/http"
+	"crypto/tls"
 
 )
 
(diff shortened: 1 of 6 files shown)


pkg/pmk/clients.go
```diff
@@ -23,7 +25,11 @@ type Client struct {
 
 // New creates the clients needed by the CLI
 // to interact with the external services.
-func NewClient(fqdn string, executor cmdexec.Executor) (Client, error) {
+func NewClient(fqdn string, executor cmdexec.Executor, allowInsecure bool) (Client, error) {
+	// Bring the hammer down to make default http allow insecure
+	if allowInsecure {
+		http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
```
sarun87, line 31: This will break if any of the clients use a custom defined `http.Transport`

Would it be better to do this per client? I mean, using `client.HttpTransport` individually for `resmgr`, `keystone` and so on rather than changing the defaultTransport.

sarun87 (approved):
See my comment inline. Looks good to me otherwise.
/lgtm

pkg/pmk/clients.go
roopakparikh: I thought about that, but seemed unnecessary, also because we may add more endpoints in future and the default http covers it all, it just means we will have every place do this when for sure we don't create new transport.

PR #29, opened 2020-12-04
CORE-1083 making segment optional - for offline installs

roopakparikh:
This makes the use of segment optional, it is controlled by
a flag. Also fixed the logging issue with the retry http module
by putting in a warpper on top of zap logger.


roopakparikh:
Since there are no reviews for past 9 days, merging it in.

PR #30, opened 2020-12-06
core-1063 Make swap permanently off by editing /etc/fstab

roopakparikh:
This code change adds an ugly sed expression to find and comment out
the swap configuration in addition to the swapoff -a command.

I am reluctantly adding it because kubernetes doesn't like swap and I am
hoping one day I will be able to convince others to let it be enabled.


roopakparikh:
Merging this in, it has been open for more than 8 days.

PR #31, opened 2020-12-14
Add redhat OS detection for old centos 7.3 image on vmware region

tlindev:


pkg/pmk/node.go
```diff
@@ -147,7 +147,7 @@ func validatePlatform(exec cmdexec.Executor) (string, error) {
 		if err != nil {
 			return "", fmt.Errorf("Couldn't read the OS configuration file os-release: %s", err.Error())
 		}
-		if strings.Contains(string(out), "7.5") || strings.Contains(string(out), "7.6") || strings.Contains(string(out), "7.7") || strings.Contains(string(out), "7.8") {
+		if strings.Contains(string(out), "7.3") || strings.Contains(string(out), "7.5") || strings.Contains(string(out), "7.6") || strings.Contains(string(out), "7.7") || strings.Contains(string(out), "7.8") {
```
roopakparikh, line 150: what about 7.4 ?

roopakparikh (changes requested):
Please add 7.4

pkg/pmk/node.go
roopakparikh: if it can't find a compatible version shouldn't it return an error  ?

roopakparikh (changes requested):

roopakparikh (approved):

roopakparikh (approved):

PR #36, opened 2021-01-12
check-node command

sahil-lakhwani:

Adds a new sub-command to check if a VM meets the prerequisites to be registered as a Kubernetes node.

Success case:
```
pf9ctl check-node
2021-01-25T16:42:48.630+0530	INFO	Loading config...
PackageCheck : PASS
SudoCheck : FAIL
CPUCheck : PASS
DiskCheck : PASS
MemoryCheck : PASS
PortCheck : PASS
```

Failure case
```

pf9ctl check-node
2021-01-25T16:42:48.630+0530	INFO	Loading config...
PackageCheck : PASS
SudoCheck : FAIL
CPUCheck : PASS
DiskCheck : PASS
MemoryCheck : PASS
PortCheck : FAIL
2021-01-25T16:42:48.903+0530	FATAL	Node not ready. See verbose logs for more
```

### DIFF
--- a/cmd/checkNode.go
+++ b/cmd/checkNode.go
@@ -1,3 +1,5 @@
+// Copyright © 2020 The pf9ctl authors
+
 package cmd
 
 import (
@@ -41,7 +43,7 @@ func checkNodeRun(cmd *cobra.Command, args []string) {
 
 	result, err := pmk.CheckNode(c)
 	if err != nil {
-		c.Segment.SendEvent("Prerequisite Node - Failed", err)
+		c.Segment.SendEvent("Prerequisite Checks for Node - Failed", err)
 		zap.S().Fatalf("Unable to check prerequisites for node: %s\n", err.Error())
 	}
 	if !result {
--- a/cmd/checkNode.go
+++ b/cmd/checkNode.go
@@ -27,7 +27,7 @@ func init() {
 }
 
 func checkNodeRun(cmd *cobra.Command, args []string) {
-	ctx, err := pmk.LoadContext(Pf9DBLoc)
+	ctx, err := pmk.LoadConfig(Pf9DBLoc)
 	if err != nil {
 		zap.S().Fatalf("Unable to load the context: %s\n", err.Error())
 	}
@@ -41,12 +41,9 @@ func checkNodeRun(cmd *cobra.Command, args []string) {
 		zap.S().Fatalf("Unable to load clients needed for the Cmd. Error: %s", err.Error())
 	}
 
-	result, err := pmk.CheckNode(c)
-	if err != nil {
-		c.Segment.SendEvent("Prerequisite Checks for Node - Failed", err)
-		zap.S().Fatalf("Unable to check prerequisites for node: %s\n", err.Error())
-	}
+	result := pmk.CheckNode(c)
+
 	if !result {
-		zap.S().Fatal("Node not ready")
+		zap.S().Fatal("Node not ready. See verbose logs for more")
 	}
 }
--- a/cmd/checkNode.go
+++ b/cmd/checkNode.go
@@ -1,0 +1,50 @@
+package cmd
+
+import (
+	"github.com/platform9/pf9ctl/pkg/pmk"
+	"github.com/spf13/cobra"
+	"go.uber.org/zap"
+)
+
+var checkNodeCmd = &cobra.Command{
+	Use:   "check-node",
+	Short: "Check prerequisites for k8s",
+	Long: `Check if a node satisfies prerequisites to be ready to be added to a Kubernetes cluster. Read more
+	at https://platform9.com/blog/support/managed-container-cloud-requirements-checklist/`,
+	Run: checkNodeRun,
+}
+
+func init() {
+	checkNodeCmd.Flags().StringVarP(&user, "user", "u", "", "ssh username for the nodes")
+	checkNodeCmd.Flags().StringVarP(&password, "password", "p", "", "ssh password for the nodes")
+	checkNodeCmd.Flags().StringVarP(&sshKey, "ssh-key", "s", "", "ssh key file for connecting to the nodes")
+	checkNodeCmd.Flags().StringSliceVarP(&ips, "ips", "i", []string{}, "ips of host to be prepared")
+	checkNodeCmd.Flags().BoolVarP(&floatingIP, "floating-ip", "f", false, "")
+
+	rootCmd.AddCommand(checkNodeCmd)
+}
+
+func checkNodeRun(cmd *cobra.Command, args []string) {
+	ctx, err := pmk.LoadContext(Pf9DBLoc)
+	if err != nil {
+		zap.S().Fatalf("Unable to load the context: %s\n", err.Error())
+	}
+
+	executor, err := getExecutor()
+	if err != nil {
+		zap.S().Fatalf("Error connecting to host %s", err.Error())
+	}
+	c, err := pmk.NewClient(ctx.Fqdn, executor, ctx.AllowInsecure, false)
+	if err != nil {
+		zap.S().Fatalf("Unable to load clients needed for the Cmd. Error: %s", err.Error())
+	}
+
+	result, err := pmk.CheckNode(c)
+	if err != nil {
+		c.Segment.SendEvent("Prerequisite Node - Failed", err)
+		zap.S().Fatalf("Unable to check prerequisites for node: %s\n", err.Error())
+	}
+	if !result {
+		zap.S().Fatal("Node not ready")
+	}
+}
--- a/pkg/pmk/checkNode.go
+++ b/pkg/pmk/checkNode.go
@@ -1,4 +1,5 @@
 // Copyright © 2020 The Platform9 Systems Inc.
+
 package pmk
 
 import (
@@ -13,10 +14,15 @@ import (
 )
 
 const (
-	minCPUs      = 2
-	minMem       = 12
-	gB           = 1024.00 * 1024
-	minDisk      = 30
+	// number of CPUs
+	minCPUs = 2
+	// RAM in GiBs
+	minMem = 12
+	// Measure of a GiB in terms of bytes
+	gB = 1024 * 1024
+	// Disk size in GiBs
+	minDisk = 30
+	// Disk size in GiBs
 	minAvailDisk = 15
 
 	checkPass = "PASS"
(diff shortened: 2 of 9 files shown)


pkg/pmk/checkNode.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+type cpuCheck struct{}
+
+func (c cpuCheck) name() string {
+	return "CPUCheck"
+}
+
+func (c cpuCheck) check(exec cmdexec.Executor) (bool, error) {
+	cpuS, err := exec.RunWithStdout("bash", "-c", "grep -c ^processor /proc/cpuinfo | tr -d '\\n'")
```
anupbarve, line 140: Can we use https://golang.org/pkg/internal/cpu/ instead.

pkg/pmk/checkNode.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+type memoryCheck struct{}
+
+func (c memoryCheck) name() string {
+	return "MemoryCheck"
+}
+
+func (c memoryCheck) check(exec cmdexec.Executor) (bool, error) {
+	memS, err := exec.RunWithStdout("bash", "-c", "echo $(($(getconf _PHYS_PAGES) * $(getconf PAGE_SIZE) / (1024 * 1024))) | tr -d '\\n'")
```
anupbarve, line 166: Can we use package like https://godoc.org/github.com/artyom/meminfo instead ?

pkg/pmk/checkNode.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+type diskCheck struct{}
+
+func (c diskCheck) name() string {
+	return "DiskCheck"
+}
+
+func (c diskCheck) check(exec cmdexec.Executor) (bool, error) {
+	diskS, err := exec.RunWithStdout("bash", "-c", "df -k . --output=size | sed 1d | xargs | tr -d '\\n'")
```
anupbarve, line 192: can we use pkg like https://godoc.org/github.com/gyuho/linux-inspect/df instead ?

pkg/pmk/checkNode.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+type portCheck struct{}
+
+func (c portCheck) name() string {
+	return "PortCheck"
+}
+
+func (c portCheck) check(exec cmdexec.Executor) (bool, error) {
+	openPorts, err := exec.RunWithStdout("bash", "-c", "netstat -tupna | awk '{print $4}' | sed -e 's/.*://' | sort | uniq")
```
anupbarve, line 234: could we use net lib to do this ?

anupbarve (commented):
I would like to avoid parsing output from the commands run using the executor (if possible).
Could you please check if we can use existing golang libraries ?
Parsing the stdout is painful and should be avoided as far as possible.
I am not sure if the existing libraries would be helpful on remote nodes though. If that is the case, then we can go ahead with the current way of implementation.

pkg/util/slices.go
```diff
@@ -0,0 +1,37 @@
+package util
+
+import (
+	"reflect"
+)
+
+func Intersect(a interface{}, b interface{}) []interface{} {
```
anupbarve, line 7: Please add documentation to the functions in this file. These being in utils, need to be easily usable by others.

pkg/pmk/checkNode.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+type osCheck struct{}
+
+func (c osCheck) name() string {
+	return "OSCheck"
+}
+
+func (c osCheck) check(exec cmdexec.Executor) (bool, error) {
+	os, err := validatePlatform(exec)
```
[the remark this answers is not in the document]
raghuP9: This is hardcoded because we want a fix set of distributions to be supported via CLI. We can move the patterns out into constants. But any support for new platform will require code change, followed by binary generation.

pkg/pmk/checkNode.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	return true, nil
+}
+
+func (c osCheck) errMessage() string {
+	return fmt.Sprint("Only Ubuntu and CentOS VMs are supported")
+}
+
+type packagesCheck struct{}
```
raghuP9, line 87: Don't add separate structs for each check in this file. Add distribution specific packages and add these methods corresponding to them as part of interface implementation or else each of these methods will morph into massive if else and switches as we keep on adding new distributions and releases.

e.g.
```
// pkg/platforms/debian/debian.go

func (plat *Debian) checkMem() (bool, error) {}

--------------------------------------------------
// pkg/platforms/centos/centos.go

func (plat *Centos) checkMem() (bool, error) {}
```

This way, you can use distribution specific bundled tools that come with OS instead of trying to find common set of commands and utilities.

Also whenever you'll add support for new distributions, you'll just have to implement the interface in a new file specific to that new distribution. This will avoid adding massive code changes in existing libraries and in turn reduce regressions.

raghuP9 (changes requested):
Added few comments. PTAL

sahil-lakhwani:
> I would like to avoid parsing output from the commands run using the executor (if possible).
> Could you please check if we can use existing golang libraries ?
> Parsing the stdout is painful and should be avoided as far as possible.
> I am not sure if the existing libraries would be helpful on remote nodes though. If that is the case, then we can go ahead with the current way of implementation.

Yes, for remote nodes, the current code executes bash commands via a SSH client. This helps in a way that same commands are run for both local and remote nodes.  I think we can keep this until we have a better approach.

pkg/pmk/checkNode.go
sahil-lakhwani: I think most the checks are same for all Linux distros. I like the idea of making platform specific implementations but that may result in duplication of code.
What I think we could do is have this checks in chain and call platform specific code only when needed (with proper interface and implementations)

Thoughts?

raghuP9: That is an assumption that most checks will always be the same. We cannot predict default packages change from vendors and then in future convolute the single codebase which in turn makes unit and functional testing of different code paths hard.

Default executable paths vary as well on different platforms so you'll end up writing the platform specific functions in a single package which could have been part of that platform module anyway. There is some duplication but it is much more maintainable in the long run. That is the very reason behind interface pattern where you can write independent implementations, extend functionality without worrying about introducing regressions to other modules.

You are also assuming that there is going to be same set of checks for all platforms. In container world, things change rapidly with the help of active community. Committing ourselves to this assumption limits our product maintenance and extendibility.

sahil-lakhwani: If the set of checks are not same for all the platforms, then I am thinking if an interface like one mentioned above will be satisfied by all the platform specific implementations.

Also that is why I am trying to have mix of both of approaches - abstracting out common implementations (like default methods in abstract classes)
cc: @anupbarve

raghuP9:
```
// pkg/platforms/debian/debian.go
package debian

type Debian struct {}

func (plat *Debian) Check() (errs []error) {

  if err := checkDebianPlat1(); err != nil {
    errs := append(errs, err)
  }

  if err := checkDebianPlat2(); err != nil {
    errs := append(errs, err)
  }

  ...
    if err := checkDebianPlatN(); err != nil {
    errs := append(errs, err)
  }

  return
}

func (plat *Debian) checkDebainPlat1() error {}
func (plat *Debian) checkDebainPlat2() error {}
func (plat *Debian) checkDebainPlat3() error {}
...
func (plat *Debian) checkDebainPlatN() error {}

--------------------------------------------------
// pkg/platforms/centos/centos.go
package centos

type Centos struct {}

func (plat *Centos) Check() (errs []error) {

  if err := checkCentosPlat1(); err != nil {
    errs := append(errs, err)
  }

  if err := checkCentosPlat2(); err != nil {
    errs := append(errs, err)
  }

  ...
    if err := checkCentosPlatN(); err != nil {
    errs := append(errs, err)
  }

  return
}

func (plat *Debian) checkCentosPlat1() error {}
func (plat *Debian) checkCentosPlat2() error {}
func (plat *Debian) checkCentosPlat3() error {}
...
func (plat *Debian) checkCentosPlatN() error {}

--------------------------------------------------
// pkg/pmk/Node.go
package node

type Platform interface {
  Check() []error
}

type Node struct {
  platform Platform
}

func NewNode() *Node {
  // check Platform using /etc/os-release file
  // then create platform specific struct
  // such as Debian{} or Centos{}

  // return &Node{Debian{}} or &Node{Cenots{}}
}

func (n *Node) Check() []errors {
  return n.platform.Check()
}
```

cmd/checkNode.go
```diff
@@ -0,0 +1,52 @@
+// Copyright © 2020 The pf9ctl authors
```
raghuP9, line 1:
```suggestion
// Copyright © 2020 The Platform9 Systems Inc.
```

pkg/pmk/checkNode.go
anupbarve: @sahil-lakhwani 
If your concern is about code duplication, IMO, it will be minor.
+1 for having distribution specific packages.

raghuP9:
@sahil-lakhwani Will you adding any unit test cases to this PR?

sahil-lakhwani:
> @sahil-lakhwani Will you be adding any unit test cases to this PR?

We'll take it in a different PR, we're deciding on a testing framework for the complete project.

Also, a note on the licensing comment -  I am not sure why that is showing as resolved - but I'll have a different PR for that too, because there are are many files which have at least 2 different license headers. So will make sure to make it uniform across.

anupbarve (approved):
Please put the console output in the description for our records.

LGTM

sahil-lakhwani:
hey @raghuP9 does this look good to you?

raghuP9:
> hey @raghuP9 does this look good to you?
Yes, can you create a JIRA ticket for unit test and post the link here?

sahil-lakhwani:
> > hey @raghuP9 does this look good to you?
> > Yes, can you create a JIRA ticket for unit test and post the link here?

This is the ticket: https://platform9.atlassian.net/browse/FT-12

anupbarve:
> > > hey @raghuP9 does this look good to you?
> > > Yes, can you create a JIRA ticket for unit test and post the link here?
> 
> This is the ticket: https://platform9.atlassian.net/browse/FT-12

Above Jira will be treated as an umbrella ticket for all the testing.
@raghuP9 The merge is blocked since you have requested changes. Could you please unblock it ?

raghuP9 (approved):

PR #37, opened 2021-01-14
get context

sahil-lakhwani:

Success case:
```
./pf9ctl context get
{"fqdn":"https://dummy-du.pf9.com","os_username":"sahil","os_password":"ZHVtbXlwYXNz","os_tenant":"service","os_region":"RegionOne","wait_period":60,"allow_insecure":false}
```

Failure case:
```
./pf9ctl context get
2021-01-15T17:56:04.889+0530	FATAL	Could not load context: stat /home/sahil/pf9/db/config.json: no such file or directory
```


pkg/util/file.go
```diff
@@ -0,0 +1,40 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		if err = file.Close(); err != nil {
+			zap.S().Error(err)
+		}
+	}()
+
+	return ioutil.ReadAll(file)
```
anupbarve, line 39: I would like to avoid putting stuff in utils dir,  if the functionality is already present in the existing libs. It seems that both the above functions will have their implementations in https://golang.org/pkg/io/ioutil/ and os libs.

anupbarve (commented):
Could you please post the usage and output of this command in the description?

anupbarve (approved):
lgtm

PR #39, opened 2021-01-21
rename context to config

sahil-lakhwani:
Renames the `context` sub-command to `config`. Also changes the related method names, variable names, etc

After the change: 

Creating config - 
```
$ pf9ctl config set
Platform9 Account URL: https://du-test-kplane-anup-3711.platform9.horse
Username: sahil@infracloud.io
Password: 
Region [RegionOne]: 
Tenant [service]: 
2021-01-21T15:10:37.980+0530	INFO	Storing config
```

Printing the config 

```
$ pf9ctl config get
{"fqdn":"https://du-test-kplane-anup-3711.platform9.horse","username":"sahil@infracloud.io","password":"c2FoaWw=","tenant":"service","region":"RegionOne","wait_period":60,"allow_insecure":false}
```

### DIFF
--- a/cmd/bootstrap.go
+++ b/cmd/bootstrap.go
@@ -4,11 +4,12 @@ package cmd
 
 import (
 	"fmt"
-	"go.uber.org/zap"
+
+	"github.com/platform9/pf9ctl/pkg/cmdexec"
 	"github.com/platform9/pf9ctl/pkg/pmk"
 	"github.com/platform9/pf9ctl/pkg/qbert"
-	"github.com/platform9/pf9ctl/pkg/cmdexec"
 	"github.com/spf13/cobra"
+	"go.uber.org/zap"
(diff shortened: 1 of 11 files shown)


anupbarve (commented):
Since we have implemented "context get", I think we should also have a corresponding "context set", which is more intuitive.
@sahil-lakhwani Could you please make that change as well as part of this PR.
All other changes look fine to me.

sahil-lakhwani:
> Since we have implemented "context get", I think we should also have a corresponding "context set", which is more intuitive.
> @sahil-lakhwani Could you please make that change as well as part of this PR.
> All other changes look fine to me.

Did this in a new commit

anupbarve (approved):
Please put the console output after the commands were run in the description for future reference before merging.

LGTM.

PR #41, opened 2021-02-02
Centos test

devidask27:
Test cases added for prerequisite check for check-node for both versions Centos and Debian

### DIFF
--- a/pkg/platform/centos/centos_test.go
+++ b/pkg/platform/centos/centos_test.go
@@ -21,6 +21,8 @@ func TestCPU(t *testing.T) {
 		args
 		want
 	}{
+		//Success case. Minimun CPU required is 2
+		//Returning 4 CPUS. Therfore test case should pass
 		"CheckPass": {
 			args: args{
 				exec: &cmdexec.MockExecutor{
@@ -33,6 +35,8 @@ func TestCPU(t *testing.T) {
 				result: true,
 			},
 		},
+		//Failure case. CPUS should be less than 2 (No of CPU < 2)
+		//Returning 1 CPUS. Therfore test case should pass
 		"CheckFail": {
 			args: args{
 				exec: &cmdexec.MockExecutor{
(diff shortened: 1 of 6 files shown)


sahil-lakhwani (changes requested):
I think we can also write tests the `err` field in the  `want` struct. For this, have the executor return an error.

pkg/platform/centos/centos_test.go
```diff
@@ -0,0 +1,323 @@
...
+			},
+			want: want{
+				result: true,
+			},
+		},
+		"CheckFail": {
+			args: args{
+				exec: &cmdexec.MockExecutor{
+					MockRunWithStdout: func(name string, args ...string) (string, error) {
+						return "1", nil
```
anupbarve, line 40: Could you please state as to why this should fail ?
From reading the test case, we wouldn't know what are success and failure conditions.
Put appropriate comments for both success and failure conditions.

This applies to all the test cases.

anupbarve:
Also run a TC build with your branch and post the TC build link here.

pkg/platform/centos/centos_test.go
devidask27: Sure i will do it.

pkg/platform/centos/centos_test.go
```diff
@@ -0,0 +1,323 @@
...
+			}
+			if diff := cmp.Diff(tc.want.result, o); diff != "" {
+				t.Errorf("r: -want, +got:\n%s", diff)
+			}
+		})
+	}
+}
+
+//Port check test case
+func TestPort(t *testing.T) {
```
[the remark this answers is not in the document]
devidask27: Tested with mixed output

pkg/platform/centos/centos_test.go
devidask27: Added comments for each test case

pkg/platform/centos/centos_test.go
```diff
@@ -21,6 +21,8 @@ func TestCPU(t *testing.T) {
 		args
 		want
 	}{
+		//Success case. Minimun CPU required is 2
+		//Returning 4 CPUS. Therfore test case should pass
```
anupbarve, line 25: nit: Therfore spelling here and everywhere.

anupbarve (changes requested):
Could you please also add test cases for `removePyCli` functionality.

PR #42, opened 2021-02-02
check and remove python cli components

sahil-lakhwani:

Checks if the any of the Python CLI components are present on the machine and removes them.

### DIFF
--- a/pkg/platform/centos/centos.go
+++ b/pkg/platform/centos/centos.go
@@ -149,3 +149,24 @@ func (c *CentOS) checkPort() (bool, error) {
 
 	return true, nil
 }
+
+func (c *CentOS) checkPyCli() (bool, error) {
+
+	_, err := c.exec.RunWithStdout("ls", util.PyCliLink)
+	if err == nil {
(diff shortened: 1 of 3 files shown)


anupbarve (changes requested):
Shouldn't these be integrated in check-node and prep-node flow ?

anupbarve:
`checkPyCli` is not called in `Check` function. Am I missing something ?

sahil-lakhwani:
> `checkPyCli` is not called in `Check` function. Am I missing something ?

yeah my bad @anupbarve . Some check-in issue I believe, have pushed a new commit.

pkg/platform/centos/centos.go
```diff
@@ -149,3 +152,24 @@ func (c *CentOS) checkPort() (bool, error) {
 
 	return true, nil
 }
+
+func (c *CentOS) checkPyCli() (bool, error) {
```
anupbarve, line 156: `checkPyCli` seems misnomer here. 
Maybe use `erasePyCli` or `removePyCli`.

anupbarve (dismissed):
nit to be addressed.
Rest looks good.
LGTM

anupbarve (approved):

PR #44, opened 2021-02-03
fix port check for remote nodes

sahil-lakhwani:

Modifies a different command for port check with remote executor.
The difference is that that while running the command with sshclient, it is wrapped by `""` hence the input to awk is to be escaped.

The executor should be refactored in future to avoid such things.

### DIFF
--- a/pkg/platform/centos/centos.go
+++ b/pkg/platform/centos/centos.go
@@ -136,7 +136,16 @@ func (c *CentOS) checkDisk() (bool, error) {
 }
 
 func (c *CentOS) checkPort() (bool, error) {
-	openPorts, err := c.exec.RunWithStdout("bash", "-c", "netstat -tupna | awk '{print $4}' | sed -e 's/.*://' | sort | uniq")
+	var arg string
+
+	switch c.exec.(type) {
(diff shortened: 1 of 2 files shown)


pkg/platform/centos/centos.go
```diff
@@ -136,7 +136,16 @@ func (c *CentOS) checkDisk() (bool, error) {
...
 
 func (c *CentOS) checkPort() (bool, error) {
-	openPorts, err := c.exec.RunWithStdout("bash", "-c", "netstat -tupna | awk '{print $4}' | sed -e 's/.*://' | sort | uniq")
+	var arg string
+
+	switch c.exec.(type) {
+	case cmdexec.LocalExecutor:
+		arg = "netstat -tupna | awk '{print $4}' | sed -e 's/.*://' | sort | uniq"
+	case *cmdexec.RemoteExecutor:
+		arg = "netstat -tupna | awk '{print \\$4}' | sed -e 's/.*://' | sort | uniq"
```
anupbarve, line 145: Could you please put a comment around this logic for future reference.

anupbarve (approved):
LGTM

anupbarve (approved):

PR #45, opened 2021-02-03
Fix common.go

devidask27:
In common.go changes did from Info to Debug


anupbarve:
you need to base this fix on master branch, so that only relevant file will be shown in the diff.
It seems to be based on your testing branch, which is why it is showing the test code as well.

PR #54, opened 2021-02-09
Test case added for PythonCli removal

devidask27:
- Changed Therefore spelling
- Added test case for PythonCli removal for both centos and debian

### DIFF
--- a/pkg/platform/centos/centos_test.go
+++ b/pkg/platform/centos/centos_test.go
@@ -21,6 +21,8 @@ func TestCPU(t *testing.T) {
 		args
 		want
 	}{
+		//Success case. Minimun CPU required is 2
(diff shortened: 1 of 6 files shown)


anupbarve (approved):

pkg/platform/centos/centos_test.go
```diff
@@ -0,0 +1,401 @@
...
+	type want struct {
+		result bool
+		err    error
+	}
+
+	cases := map[string]struct {
+		args
+		want
+	}{
+		//Success case. Packages should not be installed allready.
```
nikhilbandi, line 306: Minor typo. 'allready' should be 'already' here and all such places.

devidask27: Changes are made please review

anupbarve (approved):

PR #56, opened 2021-02-10
Prep node platform val code break

devidask27:
Modularised commands code

### DIFF
--- a/pkg/pmk/node.go
+++ b/pkg/pmk/node.go
@@ -131,42 +131,63 @@ func installHostAgentCertless(ctx Config, auth keystone.KeystoneAuth, hostOS str
 
 func validatePlatform(exec cmdexec.Executor) (string, error) {
 	zap.S().Debug("Received a call to validate platform")
-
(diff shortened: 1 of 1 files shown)


pkg/pmk/node.go
```diff
@@ -131,44 +131,66 @@ func installHostAgentCertless(ctx Config, auth keystone.KeystoneAuth, hostOS str
...
+		"cat /etc/*os-release | grep -i pretty_name | cut -d ' ' -f 2")
+	if err != nil {
+		return "", fmt.Errorf("Couldn't read the OS configuration file os-release: %s", err.Error())
+	}
+	if strings.Contains(string(out), "16") || strings.Contains(string(out), "18") {
+		return "debian", nil
+	}
+	return "", fmt.Errorf("Unable to determine OS type: %s", string(out))
+}
+
```
anupbarve, line 193: This should go in the platform specific files i.e. in pkg/platform/<centos/debian>

anupbarve (changes requested):
Also please add TC link here in the description going forward.

PR #58, opened 2021-02-11
Private/devidask/prep node testing

devidask27:
Added prep-node test cases

1. load config
2. setup node
3. get executor


devidask27:
TC build link 
http://teamcity.platform9.horse:8111/buildConfiguration/Pf9project_Platform9Components_Pf9ctl/1429929

pkg/pmk/config_test.go
```diff
@@ -0,0 +1,50 @@
...
+			args : args{
+				loc : "/tmp",
+				err : nil,
+			},
+		},
+		//Failure case. Should return error for empty file name
+		"CheckFail" : {
+			args : args{
+				loc : "",
+				err : fmt.Errorf("Config absent, run `sudo pf9ctl config set`"),
```
nikhilbandi, line 39: A better message might be, "Configuration details not provided, run ....."

pkg/pmk/config_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"fmt"
+	"github.com/stretchr/testify/assert"
+	"io/ioutil"
+	"os"
+)
+
+//Load config test case
+func TestLoadConfig(t *testing.T)  {
+
+	tmpFile, err := ioutil.TempFile("/tmp","config.json")
```
nikhilbandi, line 14: Should this path be provided in a config file from which it could be referred?

pkg/pmk/config_test.go
```diff
@@ -0,0 +1,50 @@
...
+	defer os.Remove(tmpFile.Name()+".bak")
+	if err != nil {
+		t.Errorf("Error creating a tempfile %s", err)
```
nikhilbandi, line 17: Better error message might be, "Error creating temporary file. ...."

pkg/pmk/config_test.go
devidask27: I think that is not required because we are just faking it in order to test the function

devidask27:
TC build link
http://teamcity.platform9.horse:8111/buildConfiguration/Pf9project_Platform9Components_Pf9ctl/1441137

pkg/pmk/clients_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+			if err != nil {
+				t.Errorf("Error occured : %s", err)
+			}
+			assert.Equal(t, tc.Client.Resmgr, client.Resmgr)
+			assert.Equal(t, tc.Client.Keystone, client.Keystone)
+			assert.Equal(t, tc.Client.Qbert, client.Qbert)
+			assert.Equal(t, tc.Client.Executor, client.Executor)
+			assert.Equal(t, tc.Client.Segment, client.Segment)
+		})
```
sahil-lakhwani, line 50: Should there be a failure case here?

Also, can you check if the `client` as a whole can be compared instead of individual fields in the struct .

sahil-lakhwani (changes requested):
- Please add a new line at end of every file
 - Any reason for specifically using assert.Equal instead of cmp.Diff  wherever possible - just to keep uniform code across

pkg/pmk/clients_test.go
devidask27: - cmp.Diff gives error while comparing for error thats why i used assert.Equal
- and client structure have multiple structures thats why i compared each structure separately
- and i think currently new client function  not having failure case. because every time it returns nil error

devidask27:
TC build link
http://teamcity.platform9.horse:8111/buildConfiguration/Pf9project_Platform9Components_Pf9ctl/1443210

sahil-lakhwani (approved):
LGTM

PR #60, opened 2021-02-17
better error msg

sahil-lakhwani:

### DIFF
--- a/pkg/platform/centos/centos.go
+++ b/pkg/platform/centos/centos.go
@@ -67,7 +67,6 @@ func (c *CentOS) checkSudo() (bool, error) {
 
 	id, err := strconv.Atoi(idS)
 	if err != nil {
-		zap.S().Info(">", err)
 		return false, err
 	}
 
@@ -157,7 +156,7 @@ func (c *CentOS) checkPort() (bool, error) {
 	}
 
 	openPortsArray := strings.Split(string(openPorts), "\n")
-	
+
 	intersection := util.Intersect(util.RequiredPorts, openPortsArray)
 
 	if len(intersection) != 0 {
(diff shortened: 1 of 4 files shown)


pkg/pmk/common.go
```diff
@@ -14,7 +15,7 @@ func setupNode(hostOS string, exec cmdexec.Executor) (err error) {
 		return err
 	}
 	if err := swapOffFstab(exec, "/etc/fstab"); err != nil {
-		return err
+		return fmt.Errorf("Unable to edit file /etc/fstab")
```
anupbarve, line 18: This is going to mask the actual errors and make it difficult for us to debug the failures.
Could you also add a debug log with actual err.
Same comment for other instances as well.

pkg/platform/centos/centos.go
```diff
@@ -67,7 +67,6 @@ func (c *CentOS) checkSudo() (bool, error) {
 
 	id, err := strconv.Atoi(idS)
 	if err != nil {
-		zap.S().Info(">", err)
```
anupbarve, line 70: Was wondering what this was all about while testing some stuff :) 
Thanks for removing.

anupbarve (changes requested):
Please put the console output for the manual testing.
Please also put the TC link.

pkg/pmk/common.go
sahil-lakhwani: @anupbarve The debug log for this is printed here https://github.com/platform9/pf9ctl/blob/622631834f501d061efcf78e14b59f67c1fc4f04/pkg/cmdexec/executor.go#L57

anupbarve (approved):
LGTM

PR #61, opened 2021-02-17
Added segment events to the check-node path.

anupbarve:
This patch adds segment events to the check-node code path.

Following is the snippet of an event seen in the segment debugger:

Success case:

client.Track(&analytics.Track{
  AnonymousId: "574de189-24b8-4869-a4f3-789682d68f62",
  Event: "Check Node: MemoryCheck Status: PASS",
  Properties: map[string]interface{}{
    "data": map[string]interface{}{
      "ProjectID": "1fc864503b2c4439bca59e8fa8c07be8",
      "Token": "g***-8***-K***-Z***_A***-p***",
      "UserID": "bc8afb54b3ca464bb9bf6de204d8f585",
    },
  },
})

Failure case:

client.Track(&analytics.Track{
  AnonymousId: "8e8ff06f-bd8d-489a-9128-8743c3c18316",
  Event: "Check Node: DiskCheck Status: FAIL",
  Properties: map[string]interface{}{
    "data": map[string]interface{}{
      "ProjectID": "1fc864503b2c4439bca59e8fa8c07be8",
      "Token": "g***-8***-K***-Z***_A***-p***",
      "UserID": "bc8afb54b3ca464bb9bf6de204d8f585",
    },
  },
})

Note: Here, the UserID is the keystone ID which would enable to map all these requests against a particular user.

Teamcity build link : http://teamcity.platform9.horse:8111/buildConfiguration/Pf9project_Platform9Components_Pf9ctl/1435423

### DIFF
--- a/pkg/pmk/checkNode.go
+++ b/pkg/pmk/checkNode.go
@@ -43,20 +43,20 @@ func CheckNode(ctx Config, allClients Client) (bool, error) {
 	)
 
 	if err != nil {
-		return false, fmt.Errorf("Unable to locate keystone credentials: %s", err.Error())
+		return false, fmt.Errorf("Unable to obtain keystone credentials: %s", err.Error())
 	}
 
 	checks := platform.Check()
 	result := true
 	for _, check := range checks {
 		if check.Result {
-			segment_str := "Check Node: " + check.Name + " Status: " + checkPass
+			segment_str := "CheckNode: " + check.Name + " Status: " + checkPass
 			if err := allClients.Segment.SendEvent(segment_str, auth); err != nil {
 				zap.S().Errorf("Unable to send Segment event for check node. Error: %s", err.Error())
 			}
 			fmt.Printf("%s : %s\n", check.Name, checkPass)
 		} else {
-			segment_str := "Check Node: " + check.Name + " Status: " + checkFail
+			segment_str := "CheckNode: " + check.Name + " Status: " + checkFail
 			if err := allClients.Segment.SendEvent(segment_str, auth); err != nil {
 				zap.S().Errorf("Unable to send Segment event for check node. Error: %s", err.Error())
 			}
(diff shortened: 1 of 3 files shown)


pkg/pmk/checkNode.go
```diff
@@ -34,21 +34,43 @@ func CheckNode(allClients Client) bool {
...
+	// Fetch the keystone token.
+	// This is used as a reference to the segment event.
+	auth, err := allClients.Keystone.GetAuth(
+		ctx.Username,
+		ctx.Password,
+		ctx.Tenant,
+	)
+
+	if err != nil {
+		return false, fmt.Errorf("Unable to locate keystone credentials: %s", err.Error())
```
nikhilbandi, line 46: Should we replace the word "locate" with "obtain"?

pkg/pmk/checkNode.go
```diff
@@ -34,21 +34,43 @@ func CheckNode(allClients Client) bool {
...
 	result := true
 	for _, check := range checks {
 		if check.Result {
+			segment_str := "Check Node: " + check.Name + " Status: " + checkPass
+			if err := allClients.Segment.SendEvent(segment_str, auth); err != nil {
+				zap.S().Errorf("Unable to send Segment event for check node. Error: %s", err.Error())
+			}
 			fmt.Printf("%s : %s\n", check.Name, checkPass)
 		} else {
+			segment_str := "Check Node: " + check.Name + " Status: " + checkFail
```
nikhilbandi, line 59: check.Name returns something like MemoryCheck. Should the phrase "Check Node" be mentioned as "CheckNode" for consistency?

nikhilbandi (dismissed):
Added a couple of minor comments.
Otherwise, everything looks good.

nikhilbandi (approved):
Approving.

PR #63, opened 2021-02-18
FT:21 Detect clock skew in check node

anupbarve:
set system date to an old date.

> date
Thu Apr 19 12:23:10 UTC 2012
> ./pf9ctl check-node
2012-04-19T12:20:46.9246Z	INFO	Loading config...
2012-04-19T12:20:47.0976Z	FATAL	Unable to perform pre-requisite checks on this node. Error: Possible clock skew detected. Check the system time and retry.

TC build:
http://teamcity.platform9.horse:8111/buildConfiguration/Pf9project_Platform9Components_Pf9ctl/1437445

### DIFF
--- a/pkg/pmk/checkNode.go
+++ b/pkg/pmk/checkNode.go
@@ -4,6 +4,7 @@ package pmk
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/platform9/pf9ctl/pkg/platform"
 	"github.com/platform9/pf9ctl/pkg/platform/centos"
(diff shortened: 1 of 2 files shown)


pkg/pmk/checkNode.go
```diff
@@ -43,6 +44,15 @@ func CheckNode(ctx Config, allClients Client) (bool, error) {
 	)
 
 	if err != nil {
+		// Certificate expiration is detected by the http library and
+		// only error object gets populated, which means that the http
+		// status code does not reflect the actual error code.
+		// So parsing the err to check for certificate expiration.
+		cert_expire_str := "certificate has expired or is not yet valid"
+		if strings.Contains(err.Error(), cert_expire_str) {
```
nikhilbandi, line 52: To rule out not catching of error due to case sensitivity, is it better to modify this statement like:
`if strings.Contains(strings.ToLower(err.Error()), cert_expire_str) {`    ?

pkg/pmk/checkNode.go
```diff
@@ -43,6 +44,15 @@ func CheckNode(ctx Config, allClients Client) (bool, error) {
 	)
 
 	if err != nil {
+		// Certificate expiration is detected by the http library and
+		// only error object gets populated, which means that the http
+		// status code does not reflect the actual error code.
+		// So parsing the err to check for certificate expiration.
+		cert_expire_str := "certificate has expired or is not yet valid"
```
nikhilbandi, line 51: Please see if we could store this in a constants file, if possible and available.

nikhilbandi (approved):
Changes look good to me.

PR #80, opened 2021-02-23
integrated check-node in prep-node

AnirudhPokala:
Integrated check-node in prep-node. 

Screenshot 2021-02-23 at 7 37 56 PM

### DIFF
--- a/cmd/checkNode.go
+++ b/cmd/checkNode.go
@@ -4,6 +4,7 @@ package cmd
 
 import (
 	"fmt"
+
 	"github.com/platform9/pf9ctl/pkg/log"
 	"github.com/platform9/pf9ctl/pkg/pmk"
 	"github.com/spf13/cobra"
@@ -43,6 +44,8 @@ func checkNodeRun(cmd *cobra.Command, args []string) {
 	if err != nil {
 		zap.S().Fatalf("Unable to load clients needed for the Cmd. Error: %s", err.Error())
 	}
+
+	defer c.Segment.Close()
 
 	result, err := pmk.CheckNode(ctx, c)
 	if err != nil {
--- a/cmd/prepNode.go
+++ b/cmd/prepNode.go
@@ -60,7 +60,7 @@ func prepNodeRun(cmd *cobra.Command, args []string) {
 
 	_, err = pmk.CheckNode(ctx, c)
 	if err != nil {
-		zap.S().Fatalf("Checknode - Failed: %s\n", err.Error())
+		zap.S().Fatalf("Pre-requisite check failed %s\n", err.Error())
 	}
 
 	if err := pmk.PrepNode(ctx, c); err != nil {
--- a/cmd/prepNode.go
+++ b/cmd/prepNode.go
@@ -56,6 +56,12 @@ func prepNodeRun(cmd *cobra.Command, args []string) {
 	if err != nil {
 		zap.S().Fatalf("Unable to load clients needed for the Cmd. Error: %s", err.Error())
 	}
+	defer c.Segment.Close()
+
+	_, err = pmk.CheckNode(ctx, c)
+	if err != nil {
+		zap.S().Fatalf("Checknode - Failed: %s\n", err.Error())
+	}
 
 	if err := pmk.PrepNode(ctx, c); err != nil {
 		c.Segment.SendEvent("Prep Node - Failed", err)
--- a/cmd/prepNode.go
+++ b/cmd/prepNode.go
@@ -57,15 +57,18 @@ func prepNodeRun(cmd *cobra.Command, args []string) {
 		zap.S().Fatalf("Unable to load clients needed for the Cmd. Error: %s", err.Error())
 	}
 	defer c.Segment.Close()
-
-	_, err = pmk.CheckNode(ctx, c)
+	// If all pre-requisite checks passed in Check-Node then prep-node
+	ch, err := pmk.CheckNode(ctx, c)
 	if err != nil {
 		zap.S().Fatalf("Pre-requisite check failed %s\n", err.Error())
 	}
-
-	if err := pmk.PrepNode(ctx, c); err != nil {
-		c.Segment.SendEvent("Prep Node - Failed", err)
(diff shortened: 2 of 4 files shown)


cmd/checkNode.go
```diff
@@ -44,6 +45,8 @@ func checkNodeRun(cmd *cobra.Command, args []string) {
 		zap.S().Fatalf("Unable to load clients needed for the Cmd. Error: %s", err.Error())
 	}
 
+	defer c.Segment.Close()
```
nikhilbandi, line 48: Why have we deferred the closing of Segment here and at other places?

cmd/prepNode.go
```diff
@@ -56,6 +56,12 @@ func prepNodeRun(cmd *cobra.Command, args []string) {
 	if err != nil {
 		zap.S().Fatalf("Unable to load clients needed for the Cmd. Error: %s", err.Error())
 	}
+	defer c.Segment.Close()
+
+	_, err = pmk.CheckNode(ctx, c)
+	if err != nil {
+		zap.S().Fatalf("Checknode - Failed: %s\n", err.Error())
```
nikhilbandi, line 63: Better error message is- "Pre-requisite check failed:"

cmd/checkNode.go
AnirudhPokala: Before adding differ statement I got this, since we are closing the segment in checknode.go 

Screenshot 2021-02-23 at 7 49 46 PM

cmd/prepNode.go
AnirudhPokala: sure will update it.

cmd/checkNode.go
AnirudhPokala: > Why have we deferred the closing of Segment here and at other places?

Because I have removed the segment closing part in pkg/pmk/checkNode.go, so when we will call the checknode command then its segment will not be closed. so added the differ in checknode.go also.

cmd/prepNode.go
AnirudhPokala: Done nikhil modified the error message.

pkg/pmk/checkNode.go
```diff
@@ -83,6 +83,6 @@ func CheckNode(ctx Config, allClients Client) (bool, error) {
 	}
 
 	// Segment events get posted from it's queue only after closing the client.
-	allClients.Segment.Close()
+	//allClients.Segment.Close()
```
anupbarve, line 86: Remove.

cmd/prepNode.go
```diff
@@ -56,6 +56,12 @@ func prepNodeRun(cmd *cobra.Command, args []string) {
 	if err != nil {
 		zap.S().Fatalf("Unable to load clients needed for the Cmd. Error: %s", err.Error())
 	}
+	defer c.Segment.Close()
+
+	_, err = pmk.CheckNode(ctx, c)
```
anupbarve, line 61: CheckNode is considered to fail if it returns `False`.
So we should not let the prep-node flow continue in 2 cases.
- If an error is encountered. (This you have covered.)
- If checkNode returns False. (This needs to be implemented.)

anupbarve (commented):
We should also remove the following client close call in prep-node flow to avoid multiple closures.
https://github.com/platform9/pf9ctl/blob/master/pkg/pmk/node.go#L24

anupbarve (changes requested):
Please also put the failure case in the description, where user runs prep-node command and checkNode fails.

pkg/pmk/checkNode.go
AnirudhPokala: Sure anup.

AnirudhPokala:
> Please also put the failure case in the description, where user runs prep-node command and checkNode fails.

Screenshot 2021-02-24 at 1 16 21 PM

pkg/pmk/node.go
```diff
@@ -21,7 +22,7 @@ func sendSegmentEvent(allClients Client, eventStr string, auth keystone.Keystone
 		zap.S().Errorf("Unable to send Segment event for Node prep. Error: %s", err.Error())
 	}
 	// Segment events get posted from it's queue only after closing the client.
-	allClients.Segment.Close()
+	//allClients.Segment.Close()
```
anupbarve, line 25: Please delete these.

pkg/pmk/checkNode.go
anupbarve: Please delete this instead of commenting-out.

cmd/prepNode.go
```diff
@@ -56,10 +56,19 @@ func prepNodeRun(cmd *cobra.Command, args []string) {
...
+	if err != nil {
+		zap.S().Fatalf("Pre-requisite check failed %s\n", err.Error())
+	}
+	if ch {
+		if err := pmk.PrepNode(ctx, c); err != nil {
+			c.Segment.SendEvent("Prep Node - Failed", err)
+			zap.S().Fatalf("Unable to prep node: %s\n", err.Error())
+		}
+	} else {
+		zap.S().Fatalf("Check node - Failed\n")
```
anupbarve, line 71: Make this:
Pre-requisite check(s) failed.

cmd/prepNode.go
```diff
@@ -56,10 +56,19 @@ func prepNodeRun(cmd *cobra.Command, args []string) {
...
 	}
-
-	if err := pmk.PrepNode(ctx, c); err != nil {
-		c.Segment.SendEvent("Prep Node - Failed", err)
-		zap.S().Fatalf("Unable to prep node: %s\n", err.Error())
+	defer c.Segment.Close()
+	// If all pre-requisite checks passed in Check-Node then prep-node
+	ch, err := pmk.CheckNode(ctx, c)
+	if err != nil {
+		zap.S().Fatalf("Pre-requisite check failed %s\n", err.Error())
```
anupbarve, line 63: s/check/check(s)

anupbarve (changes requested):
Minor comments. Will approve once addressed.

anupbarve (approved):

PR #85, opened 2021-02-24
run all commands with sudo after checking for sudo

sahil-lakhwani:

Changes the LocalExecutor to run all commands with sudo
Checks if the current user is ableto sudo.

Have tested this with Ubuntu and CentOS

When running with non sudoer
```
./pf9ctl check-node
2021-02-24T11:22:26.2234Z	INFO	Loading configuration details
[sudo] password for dummy1: 
2021-02-24T11:22:28.6002Z	FATAL	Unable to perform pre-requisite checks on this node: User executing this CLI is not allowed to switch to privileged (sudo) mode
```

### DIFF
--- a/pkg/pmk/node.go
+++ b/pkg/pmk/node.go
@@ -3,6 +3,10 @@ package pmk
 
 import (
 	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
 	"github.com/platform9/pf9ctl/pkg/cmdexec"
 	"github.com/platform9/pf9ctl/pkg/keystone"
 	"github.com/platform9/pf9ctl/pkg/platform"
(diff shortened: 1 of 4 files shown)


pkg/pmk/node.go
```diff
@@ -234,3 +235,8 @@ func installHostAgentLegacy(ctx Config, auth keystone.KeystoneAuth, hostOS strin
 	zap.S().Info("Hostagent installed successfully")
 	return nil
 }
+
+func checkSudo(exec cmdexec.Executor) bool {
+	_, err := exec.RunWithStdout("bash", "-c", "echo '$(logname) ALL=(ALL:ALL) NOPASSWD: ALL' | (EDITOR='tee -a' visudo)")
```
anupbarve, line 240: What is `logname` ? I cannot find it's reference.

sahil-lakhwani: It's the login name I believe, the result of the command on a  machine with ubuntu user would be `sudo bash -c 'echo "$(logname) ALL=(ALL:ALL) NOPASSWD: ALL" | (EDITOR="tee -a" visudo)'`

But I think running `sudo -l` has the same behaviour and less chances of failure. Let me check and update the code if that is the case

sahil-lakhwani: Have update this with `sudo -l`

anupbarve (approved):
LGTM

PR #89, opened 2021-02-25
Prompted the config if not set, moved all constants to util constants.go

AnirudhPokala:
* Created ConfigCmdCreateRun() function in config.go (pmk).
* Prompted the config if not set (during check-node/ prep-node) 
* Moved the constants in cmd/constants.go to pkg/util/constants.go 
Screenshot 2021-02-25 at 1 46 04 PM


pkg/util/constants.go
```diff
@@ -47,9 +48,28 @@ var (
...
 
 func init() {
 	RequiredPorts = []string{"443", "2379", "2380", "8285", "10250", "10255", "4194", "8285"}
 	PortErr = fmt.Sprintf("Ports required to be available %s", RequiredPorts)
 }
+
+//This is the current version of cli.
+const (
+	Version string = "pf9ctl version: v1.0"
+)
```
anupbarve, line 75: Why is this visible in the current diff ?
Can you rebase and repost ?

AnirudhPokala: we have moved all variables and constants from cmd/constants.go to pkg/util/constants.go right.

pkg/pmk/config.go
```diff
@@ -23,6 +28,23 @@ type Config struct {
...
+	Pf9DBDir = filepath.Join(Pf9Dir, "db")
+	// Pf9DBLoc represents location of the config file.
+	Pf9DBLoc = filepath.Join(Pf9DBDir, "config.json")
+	// Pf9Log represents location of the log.
+	Pf9Log = filepath.Join(Pf9LogDir, "pf9ctl.log")
+	// WaitPeriod is the sleep period for the cli
+	// before it starts with the operations.
+	WaitPeriod = time.Duration(60)
+)*/
+
```
anupbarve, line 47: Remove.

pkg/pmk/config.go
```diff
@@ -47,7 +70,10 @@ func LoadConfig(loc string) (Config, error) {
 	if err != nil {
 
 		if os.IsNotExist(err) {
-			return Config{}, ErrConfigurationDetailsNotProvided
+			// to initiate the config create and store it
```
anupbarve, line 73: Could you please add a console log saying that existing config not found, prompting for new config.

anupbarve (changes requested):

pkg/pmk/config.go
anupbarve: Hey @AnirudhPokala , 
I cannot see this change in new commit.

anupbarve (approved):

PR #90, opened 2021-02-25
auto install OS packages

sahil-lakhwani:

Automatically install missing OS packages

### DIFF
--- a/pkg/platform/centos/centos.go
+++ b/pkg/platform/centos/centos.go
@@ -14,7 +14,8 @@ import (
 )
 
 var (
-	packages = [2]string{"ntp", "curl"}
+	packages            = []string{"curl"}
+	packageInstallError = "Packages not found and could not be installed"
 )
 
 // CentOS reprents centos based host machine
@@ -70,12 +71,16 @@ func (c *CentOS) checkExistingInstallation() (bool, error) {
 
 func (c *CentOS) checkOSPackages() (bool, error) {
 
-	errLines := []string{"Packages not found: "}
+	errLines := []string{packageInstallError}
 
 	for _, p := range packages {
 		err := c.exec.Run("bash", "-c", fmt.Sprintf("yum list installed %s", p))
 		if err != nil {
-			errLines = append(errLines, p)
+			zap.S().Debugf("Package %s not found, trying to install", p)
+			if err = c.installOSPackages(p); err != nil {
+				errLines = append(errLines, p)
+			}
+			zap.S().Infof("Missing package %s installed", p)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/platform/centos/centos.go
```diff
@@ -14,7 +14,8 @@ import (
 )
 
 var (
-	packages = [2]string{"ntp", "curl"}
+	packages            = []string{"curl"}
```
anupbarve, line 17: why remove ntp ?

pkg/platform/centos/centos.go
```diff
@@ -229,3 +234,15 @@ func (c *CentOS) Version() (string, error) {
 	return "", fmt.Errorf("Unable to determine OS type: %s", string(out))
 
 }
+
+func (c *CentOS) installOSPackages(p string) error {
+	zap.S().Debug("Trying apt update...")
+	_, err := c.exec.RunWithStdout("bash", "-c", "yum update -y")
+	if err != nil {
+		return err
+	}
```
anupbarve, line 243: Wouldn't this be called in a loop at line 76 above, if there are multiple packages to be installed ? Same for debian.

pkg/platform/centos/centos.go
sahil-lakhwani: I have noticed that ntp is not needed on centos for our requirement, I could run prep-node successfully.

anupbarve: I believe it is required by out installer. We should keep it.

nikhilbandi: Yeah, as they say if something is working, don't touch it :)

pkg/platform/centos/centos.go
```diff
@@ -14,7 +14,8 @@ import (
 )
 
 var (
-	packages = [2]string{"ntp", "curl"}
+	packages            = []string{"curl"}
+	packageInstallError = "Packages not found and could not be installed"
```
nikhilbandi, line 18: Are we also mentioning the reason as to why it could not be installed.

pkg/platform/debian/debian.go
```diff
@@ -13,7 +13,8 @@ import (
 )
 
 var (
-	packages = [2]string{"ntp", "curl"}
+	packages            = []string{"ntp", "curl"}
```
nikhilbandi, line 16: Didn't get why are we hardcoding just 2 packages here. If we are going to call the function installOSPackages for all the missing packages, wouldn't it automatically detect and install all the missing packages? May be I am missing something here.

pkg/platform/debian/debian.go
```diff
@@ -228,3 +233,15 @@ func (d *Debian) Version() (string, error) {
 	}
 	return "", fmt.Errorf("Unable to determine OS type: %s", string(out))
 }
+
+func (d *Debian) installOSPackages(p string) error {
+	zap.S().Debug("Trying apt update...")
+	_, err := d.exec.RunWithStdout("bash", "-c", "apt update")
```
nikhilbandi, line 239: Should we also check if apt / yum themselves are installed or not?

pkg/platform/centos/centos.go
sahil-lakhwani: Added a debug log

pkg/platform/debian/debian.go
sahil-lakhwani: If they are not, error is logged and the user should most likely try to install themselves.

pkg/platform/debian/debian.go
sahil-lakhwani: We're selectively checking and installing the packages which are identified as necessary for this CLI to run. 
We may not be able to check that dynamically as there may be some libraries used offered by different packages.
Those cases have to identified and this list has to be updated accordingly

anupbarve (approved):

nikhilbandi (approved):
Looks good. Hope the Teamcity build has passed.

PR #92, opened 2021-02-26
log checking os packages message

sahil-lakhwani:


pkg/platform/centos/centos.go
```diff
@@ -72,6 +72,7 @@ func (c *CentOS) checkExistingInstallation() (bool, error) {
 func (c *CentOS) checkOSPackages() (bool, error) {
 
 	errLines := []string{packageInstallError}
+	zap.S().Info("Checking OS Packages")
```
anupbarve, line 75: Should we also add a separate msg for installing as well ?

Installing missing OS Packages (This might take few minutes)

sahil-lakhwani: Added a message

anupbarve (approved):

PR #93, opened 2021-02-26
Add 3306, fix console messaging

anupbarve:


anupbarve:
# bin/pf9ctl check-node
2021-02-26T09:00:31.9776Z	INFO	Loading configuration details


Removal of existing CLI : PASS
Existing Installation Check : PASS
OS Packages Check : PASS
SudoCheck : PASS
CPUCheck : PASS
DiskCheck : FAIL - At least 30 GB of total disk space and 15 GB of free space is needed on host. Disk Space found: 20 GB
MemoryCheck : PASS
PortCheck : FAIL - Following ports should not be in use: [3306]

Pre-requisite checks failed. See /root/pf9/log/pf9ctl-20210222.log or use --verbose for logs




TC build: http://teamcity.platform9.horse:8111/buildConfiguration/Pf9project_Platform9Components_Pf9ctl/1449490

pkg/platform/centos/centos.go
```diff
@@ -179,7 +179,7 @@ func (c *CentOS) checkDisk() (bool, error) {
 	return false, fmt.Errorf("Available disk space: %.0f GB", math.Trunc(avail/util.GB))
 }
 
-func (c *CentOS) checkPort() (bool, error) {
+func (c *CentOS) checkPort() (bool, string, error) {
```
sahil-lakhwani, line 182: Wondering if we can use the `err` itself to return the list of ports which are intersecting, because the intersecting list is just to be printed and not for any other use

anupbarve: done.

nikhilbandi (approved):
Changes look good.

PR #95, opened 2021-02-26
added os check in pf9ctl_setup

devidask27:
Screenshot 2021-02-26 at 4 09 41 PM

### DIFF
--- a/pf9ctl_setup
+++ b/pf9ctl_setup
@@ -97,9 +97,9 @@ EOF
 
 checkOS_version() {
     cat /etc/*release &>/dev/null
-    if [ $? -eq 1 ];
+    if [ $? != 0 ];
     then
-        echo "This OS is not supported"
+        echo "This OS is not supported."
         exit
     fi
 }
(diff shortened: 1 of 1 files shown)


devidask27:
TC build link
http://teamcity.platform9.horse:8111/buildConfiguration/Pf9project_Platform9Components_Pf9ctl/1449504

pf9ctl_setup
```diff
@@ -95,6 +95,15 @@ cat << "EOF"
 EOF
 }
 
+checkOS_version() {
+    cat /etc/*release &>/dev/null
+    if [ $? -eq 1 ];
```
anupbarve, line 100: Should check for not equal to 0 instead.

pf9ctl_setup
```diff
@@ -95,6 +95,15 @@ cat << "EOF"
...
+    then
+        echo "This OS is not supported"
```
anupbarve, line 102: Add a . at the end.

anupbarve (changes requested):

pf9ctl_setup
devidask27: Done

pf9ctl_setup
devidask27: Done

anupbarve (approved):

PR #96, opened 2021-02-26
Fix error append

anupbarve:


nikhilbandi (approved):
Looks good. Only err object with sPrintf() should be implemented.

PR #100, opened 2021-02-27
log file name fix

devidask27:
consol output
root@devidas-test-ubuntu16:/home/ubuntu/pf9/log# ls
pf9ctl-20210227.log   pf9ctl-2021102.log

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -12,7 +12,7 @@ import (
 
 // Returns the current log file location.
 func GetLogLocation(logFile string) string {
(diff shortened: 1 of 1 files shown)


devidask27:
Output
root@devidas-test-ubuntu16:/home/ubuntu/pf9/log# ls
pf9ctl-20210203.log    pf9ctl-20210227.log

anupbarve (approved):

nikhilbandi (commented):
Change looks good.
Please test after setting machine date to some different value other than today and make sure the log file with same date gets created.

nikhilbandi (commented):
Also, please make sure for different dates, different log files get generated even though there are log files present for earlier dates. Do check if the logs for a particular date are written on the file with the same date only.

devidask27:
> Change looks good.
> Please test after setting machine date to some different value other than today and make sure the log file with same date gets created.

yes i tested this. different log file is getting created for different date.

PR #101, opened 2021-03-01
message modified

devidask27:


pf9ctl_setup
```diff
@@ -100,7 +100,7 @@ checkOS_version() {
     cat /etc/*release &>/dev/null
     if [ $? != 0 ];
     then
-        echo "This OS is not supported."
+        echo "This OS is not supported. Supported operating systems are: Ubuntu (16.04, 18.04) & CentOS (7.6)."
```
sahil-lakhwani, line 103: Can we confirm that only mentioned version of CentOS is supported and no other?

anupbarve: I checked in our docs as well did some testing with centos7.8. We support centos7.x.

devidask27: Yes we support 7.8 also. Anup tested for 7.8 it works fine. I will change that message to 7.X

PR #104, opened 2021-03-01
Created a function "validateUserCredentials()" to validate the user credentials during config set, check-node, prep-node execution.

AnirudhPokala:
Screenshot 2021-03-01 at 6 58 58 PM
Screenshot 2021-03-01 at 6 58 17 PM

### DIFF
--- a/cmd/checkNode.go
+++ b/cmd/checkNode.go
@@ -80,7 +80,7 @@ func validateUserCredentials(pmk.Config) pmk.Client {
 	)
 
 	if err != nil {
-		zap.S().Fatalf("Invalid Credentials (Username/Password/Service), run 'pf9ctl config set' !!")
+		zap.S().Fatalf("Invalid credentials (Username/ Password/ Account), run 'pf9ctl config set' with correct credentials.")
 	}
 	return c
 }
--- a/cmd/checkNode.go
+++ b/cmd/checkNode.go
@@ -20,6 +20,12 @@ var checkNodeCmd = &cobra.Command{
 	Run: checkNodeRun,
 }
 
+var (
+	ctx pmk.Config
+	err error
+	c   pmk.Client
+)
+
 func init() {
 	checkNodeCmd.Flags().StringVarP(&user, "user", "u", "", "ssh username for the nodes")
 	checkNodeCmd.Flags().StringVarP(&password, "password", "p", "", "ssh password for the nodes")
@@ -32,19 +38,12 @@ func init() {
 
 func checkNodeRun(cmd *cobra.Command, args []string) {
 	zap.S().Debug("==========Running check-node==========")
-	ctx, err := pmk.LoadConfig(util.Pf9DBLoc)
+	ctx, err = pmk.LoadConfig(util.Pf9DBLoc)
 	if err != nil {
 		zap.S().Fatalf("Unable to load the context: %s\n", err.Error())
 	}
-
-	executor, err := getExecutor()
-	if err != nil {
-		zap.S().Fatalf("Error connecting to host %s", err.Error())
-	}
-	c, err := pmk.NewClient(ctx.Fqdn, executor, ctx.AllowInsecure, false)
-	if err != nil {
-		zap.S().Fatalf("Unable to load clients needed for the Cmd. Error: %s", err.Error())
-	}
+	// Validate the user credentials entered during config set and will bail out if invalid
+	c = validateUserCredentials(ctx)
 
 	defer c.Segment.Close()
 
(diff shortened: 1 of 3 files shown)


cmd/checkNode.go
```diff
@@ -61,3 +60,27 @@ func checkNodeRun(cmd *cobra.Command, args []string) {
...
+	}
+
+	_, err = c.Keystone.GetAuth(
+		ctx.Username,
+		ctx.Password,
+		ctx.Tenant,
+	)
+
+	if err != nil {
+		zap.S().Fatalf("Invalid Credentials (Username/Password/Service), run 'pf9ctl config set' !!")
```
nikhilbandi, line 83: Better error message is: "Invalid credentials (Username/ Password/ Account), run 'pf9ctl config set' with correct credentials. "

AnirudhPokala: Done Nikhil edited the error message.

cmd/checkNode.go
```diff
@@ -32,19 +38,12 @@ func init() {
...
-	executor, err := getExecutor()
-	if err != nil {
-		zap.S().Fatalf("Error connecting to host %s", err.Error())
-	}
-	c, err := pmk.NewClient(ctx.Fqdn, executor, ctx.AllowInsecure, false)
-	if err != nil {
-		zap.S().Fatalf("Unable to load clients needed for the Cmd. Error: %s", err.Error())
-	}
+	// Validate the user credentials entered during config set and will bail out if invalid
+	c = validateUserCredentials(ctx)
```
anupbarve, line 46: I don't like the idea of a validation function returning a client and operation flow using it. It is non-intuitive. We should pass the client to the `validateUserCredentials`.

AnirudhPokala: Ok anup will create a seperate function to create the client and will pass it to validateUserCredentials. Is it ok?

cmd/checkNode.go
```diff
@@ -61,3 +63,32 @@ func checkNodeRun(cmd *cobra.Command, args []string) {
...
+	_, err = c.Keystone.GetAuth(
+		ctx.Username,
+		ctx.Password,
+		ctx.Tenant,
+	)
+
+	if err != nil {
+		zap.S().Fatalf("Invalid credentials (Username/ Password/ Account), run 'pf9ctl config set' with correct credentials.")
+	}
+}
```
anupbarve, line 79: We should move this to config.go

cmd/checkNode.go
```diff
@@ -61,3 +63,32 @@ func checkNodeRun(cmd *cobra.Command, args []string) {
...
+	if err != nil {
+		zap.S().Fatalf("Error connecting to host %s", err.Error())
+	}
+	c, err := pmk.NewClient(ctx.Fqdn, executor, ctx.AllowInsecure, false)
+	if err != nil {
+		zap.S().Fatalf("Unable to load clients needed for the Cmd. Error: %s", err.Error())
+	}
+
+	return c
+}
```
anupbarve, line 94: This should not be a different function at all. We should simply unfold it and call it in respective cmd/<command.go> files.

anupbarve (changes requested):

cmd/checkNode.go
AnirudhPokala: sure anup done.

cmd/checkNode.go
AnirudhPokala: ok anup done once review it.

anupbarve (approved):
Please take help from @devidask27 to run TC build and post the TC link here.

cmd/config.go
```diff
@@ -72,3 +97,17 @@ func init() {
 	configCmdCreate.AddCommand(configCmdGet)
 	configCmdCreate.AddCommand(configCmdSet)
 }
+
+// This function will validate the user credentials entered during config set and bail out if invalid
+func validateUserCredentials(pmk.Config, pmk.Client) {
```
sahil-lakhwani, line 102: I think this function should return error instead of failing in a fatal way.

AnirudhPokala: Ya we can do that, but then we are adding same error message line thirce in config.go, checkNode.go, prepNode.go. So added that error validation in function itself.

AnirudhPokala:
> Please take help from @devidask27 to run TC build and post the TC link here.

Ya Anup its working fine. 

http://teamcity.platform9.horse:8111/buildConfiguration/Pf9project_Platform9Components_Pf9ctl/1452392

cmd/config.go
sahil-lakhwani: I think that should be fine. This function can return an error message, which the callers can use however they want

AnirudhPokala: Ya Sahil modified it, once review it.

sahil-lakhwani (approved):

PR #112, opened 2021-03-03
Resolved the issues found during testing.

AnirudhPokala:
1) Missing packages being installed during check-node/prep-node execution for remote and local case for uuid-runtime in centos was not getting installed using "yum -q -y install uuid-runtime", so modified it to uuid for centos case. 

2) Modified extra console error message to **debug** type which were found during checknode and prep-node execution when os packages were missing (remote).

3) Modified the error message we get while we enter invalid Username/Password/IP during remote running of commands. 

TC link : http://teamcity.platform9.horse:8111/buildConfiguration/Pf9project_Platform9Components_Pf9ctl/1454163

Screenshot 2021-03-03 at 2 46 30 PM

Screenshot 2021-03-03 at 2 48 51 PM

Screenshot 2021-03-03 at 2 50 41 PM

Screenshot 2021-03-03 at 2 51 34 PM

### DIFF
--- a/cmd/checkNode.go
+++ b/cmd/checkNode.go
@@ -40,7 +40,8 @@ func checkNodeRun(cmd *cobra.Command, args []string) {
 
 	executor, err := getExecutor()
 	if err != nil {
-		zap.S().Fatalf("Error connecting to host %s", err.Error())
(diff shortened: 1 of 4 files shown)


anupbarve (approved):
lgtm

PR #114, opened 2021-03-08
Config is loop backed if credentials are invalid during config set command (config/check-node/prep-node).

AnirudhPokala:
* Added the loop back part for check-node, config, prep-node if user enters invalid credentials during config set. 
Screenshot 2021-03-08 at 6 30 55 PM

Screenshot 2021-03-08 at 6 35 12 PM

Screenshot 2021-03-08 at 4 29 47 PM

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -24,34 +24,43 @@ var (
 	ctx pmk.Config
 	err error
 	c   pmk.Client
+	// This flag is used to loop back if user enters invalid credentials during config set.
+	credentialFlag bool
 )
 
 func configCmdCreateRun(cmd *cobra.Command, args []string) {
 	zap.S().Debug("==========Running set config==========")
+	credentialFlag = true
+	for credentialFlag {
+		// invoked the configcreate command from pkg/pmk
+		ctx, _ = pmk.ConfigCmdCreateRun()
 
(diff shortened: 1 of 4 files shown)


cmd/config.go
```diff
@@ -24,34 +24,43 @@ var (
...
 	err error
 	c   pmk.Client
+	// This flag is used to loop back if user enters invalid credentials during config set.
+	credentialFlag bool
 )
 
 func configCmdCreateRun(cmd *cobra.Command, args []string) {
 	zap.S().Debug("==========Running set config==========")
+	credentialFlag = true
+	for credentialFlag {
```
anupbarve, line 34: Where is this `credentialFlag` toggled in below logic ?
Am I missing something?

AnirudhPokala: We should be able to set config any time we can right, and will only loop back if user enters invalid credentials, so just break loop when we have valid credentials and then we will store it.

AnirudhPokala: we can even set credentaialsFlag = false, instead of breaking both will work.

anupbarve: In that case, just remove this flag. It is making the code reading confusing.

anupbarve (approved):

PR #115, opened 2021-03-11
Added environment variable to control segment events

devidask27:


anupbarve (commented):
Could you please add the purpose of this environment variable in README.md as well ?

devidask27:
> Could you please add the purpose of this environment variable in README.md as well ?

Done

anupbarve (dismissed):

pkg/pmk/segment.go
```diff
@@ -28,7 +29,9 @@ type NoopSegment struct {
 
 func NewSegment(fqdn string, noTracking bool) Segment {
 	// mock out segment if the user wants no Tracking
-	if noTracking {
+	envCheck := os.Getenv("PF9CTL_SEGMENT_EVENTS_DISABLE")
+	segmentEventDisabled , _:= strconv.ParseBool(envCheck)
+	if noTracking || segmentEventDisabled {
```
nikhilbandi, line 34: Just a curious question. In which scenario(s) is the bool noTracking set to True?

anupbarve: It is a functionality to provide more granularity for different commands wrt posting segment events.
Not used at this moment.

devidask27: For now it is always false

anupbarve (approved):
Please broadcast within our team on how to use this and also encourage them to use it if not working with segment events.

PR #122, opened 2021-03-17
Stoped out after 3 recursive loops during loopback config state.

AnirudhPokala:
Screenshot 2021-03-17 at 4 13 41 PM

Screenshot 2021-03-17 at 4 15 45 PM


anupbarve (changes requested):
I think the logic should be :

- If the config is invalid after reading the file, prompt for new config
- If user enters invalid config for 3 times, stop the process.

nikhilbandi:
> I think the logic should be :
> 
> * If the config is invalid after reading the file, prompt for new config
> * If user enters invalid config for 3 times, stop the process.

Agree with Anup's remark. Current logic of looping 3 times without giving a chance to user to enter correct details, is incorrect.

PR #144, opened 2021-04-22
injected segment key

AnirudhPokala:
Segment key injected

### DIFF
--- a/pkg/pmk/segment.go
+++ b/pkg/pmk/segment.go
@@ -16,8 +16,6 @@ import (
 //Added segment key for the source PRD-PMKFT Metrics-Aggregator
 var SegmentWriteKey string
 
-//var SegmentWriteKey = os.Getenv("SEGMENT_KEY_PRD_PMKFT")
-
 type Segment interface {
 	SendEvent(string, interface{}, string, string) error
 	SendGroupTraits(string, interface{}) error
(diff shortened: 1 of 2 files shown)


nikhilbandi (changes requested):
Please handle the Else case as suggested by Anup.

pkg/pmk/segment.go
```diff
@@ -14,7 +14,10 @@ import (
 )
 
 //Added segment key for the source PRD-PMKFT Metrics-Aggregator
-const segmentWriteKey = "4jevYUNBF5sY3vZJWm5TrhfdsFzQIQ3y"
+//const segmentWriteKey = "4jevYUNBF5sY3vZJWm5TrhfdsFzQIQ3y"
+var SegmentWriteKey string
+
+//var SegmentWriteKey = os.Getenv("SEGMENT_KEY_PRD_PMKFT")
```
ParagDoke, line 19: Please drop the comment.

pkg/pmk/segment.go
```diff
@@ -34,10 +37,16 @@ func NewSegment(fqdn string, noTracking bool) Segment {
 	// mock out segment if the user wants no Tracking
 	envCheck := os.Getenv("PF9CTL_SEGMENT_EVENTS_DISABLE")
 	segmentEventDisabled, _ := strconv.ParseBool(envCheck)
+
+	// Local Case.
```
ParagDoke, line 40: Comment should read "Local build case" to unambiguously capture if the build is local or run is local.

ParagDoke (commented):
Must squash commits before merging. Currently there are 11 of them.

pkg/pmk/segment.go
AnirudhPokala: Sure dropped off.

pkg/pmk/segment.go
AnirudhPokala: Sure will modify.

anupbarve (approved):

devidask27 (approved):

nikhilbandi (approved):
Looks good.

PR #151, opened 2021-05-07
Added supportBundle test file

ShaunakJoshi1407:
>Added unit test cases for supportBundle
>Modified the S3Upload and the RemoveBundle functions in supportBundle.go for ease in testing.
>Added additional comments in supportBundle.go

### DIFF
--- a/cmd/supportBundle_test.go
+++ b/cmd/supportBundle_test.go
@@ -8,7 +8,7 @@ import (
 
 	"github.com/platform9/pf9ctl/pkg/cmdexec"
 	"github.com/platform9/pf9ctl/pkg/supportBundle"
-	"github.com/stretchr/testify/assert"
+	"github.com/platform9/pf9ctl/pkg/test_utils"
 )
 
 //Errors returned by the functions
@@ -64,7 +64,7 @@ func TestHostIP(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 
 			_, err1 := supportBundle.HostIP(tc.exec)
-			assert.Equal(t, tc.want.err, err1)
+			test_utils.Equals(t, tc.want.err, err1)
 		})
 	}
 
@@ -120,7 +120,7 @@ func TestGenSupportBundle(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 
 			_, err1 := supportBundle.GenSupportBundle(tc.exec, timestamp)
-			assert.Equal(t, tc.want.err, err1)
+			test_utils.Equals(t, tc.want.err, err1)
 		})
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/supportBundle/supportBundle.go
```diff
@@ -59,8 +64,9 @@ func SupportBundleUpload(ctx pmk.Config, allClients pmk.Client) error {
 		zap.S().Debugf(color.Red("x ")+"Failed to generate supportBundle\n", err.Error())
 	}
 
+	//var exec cmdexec.Executor
```
devidask27, line 67: Remove this

pkg/supportBundle/supportBundle.go
```diff
@@ -123,16 +127,16 @@ func SupportBundleUpload(ctx pmk.Config, allClients pmk.Client) error {
 	}
 
 	// Remove the supportbundle after uploading to S3
-	errremove := allClients.Executor.Run("bash", "-c", fmt.Sprintf("rm -rf %s", fileloc))
+	errremove := RemoveBundle(allClients.Executor)
 	if errremove != nil {
-		zap.S().Debug("Failed to remove supportbundle", errremove)
+		zap.S().Debug("Error removing generated bundle from local machine", errremove)
```
AnirudhPokala, line 132: We will even remove generated bundle in remote case also right?

pkg/supportBundle/supportBundle.go
```diff
@@ -147,61 +151,79 @@ func genTargetFilename(timestamp time.Time, hostname string) string {
...
+		return ErrRemove
+	}
+	return nil
+}
+
 //This function is used to generate the support bundles.
 //It copies all the log files specified into a directory and archives that given directory.
-
-func genSupportBundle(allClients pmk.Client, timestamp time.Time) (string, error) {
+func GenSupportBundle(exec cmdexec.Executor, timestamp time.Time) (string, error) {
```
devidask27, line 175: This function executes multiple commands. and we are only testing it for nil error. i think we should add failure cases as well.

pkg/supportBundle/supportBundle.go
ShaunakJoshi1407: Ok will do

pkg/supportBundle/supportBundle.go
ShaunakJoshi1407: Yes it will get removed even in the remote case

ShaunakJoshi1407: I will change this particular message

cmd/supportBundle_test.go
```diff
@@ -0,0 +1,268 @@
...
+				err: ErrRemove,
+			},
+		},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+
+			err1 := supportBundle.RemoveBundle(tc.exec)
+			assert.Equal(t, tc.want.err, err1)
```
anupbarve, line 264: Could you reuse this implementation instead? https://github.com/platform9/pf9ctl/blob/master/pkg/test_utils/test_utils.go#L12
This will keep things uniform all across the code.

Please change this all across the testing code.

anupbarve (changes requested):

cmd/supportBundle_test.go
ShaunakJoshi1407: Sure will make the changes

anupbarve (approved):

PR #153, opened 2021-05-12
Adding log statement to include the hostname on which a particular op…

roopakparikh:
Added a log statement to make it easy for customers to debug which host failed in a multi-threaded environment.

I am still testing it end to end, sending it out for a quick review

### DIFF
--- a/pkg/cmdexec/executor.go
+++ b/pkg/cmdexec/executor.go
@@ -42,6 +42,8 @@ func (c LocalExecutor) RunWithStdout(name string, args ...string) (string, error
 // RemoteExecutor as the name implies runs commands usign SSH on remote host
 type RemoteExecutor struct {
 	Client ssh.Client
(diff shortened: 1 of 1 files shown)


pkg/cmdexec/executor.go
```diff
@@ -42,6 +42,8 @@ func (c LocalExecutor) RunWithStdout(name string, args ...string) (string, error
 // RemoteExecutor as the name implies runs commands usign SSH on remote host
 type RemoteExecutor struct {
 	Client ssh.Client
+	// host name for debugging purposes
```
anupbarve, line 45: nit :
This is generic enough and can be used to log info other than host name. I am anticipating this functionality to get updated for ssh tunneling as well and we may want to log that info too.
Maybe just update the comment to indicate so ?

anupbarve (approved):

PR #154, opened 2021-05-18
Updated README.md

ShaunakJoshi1407:
-Added requirements of pf9ctl.
-Added usage along with description of the commands in use.
-Explained each command with a sample code snippet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,266 @@
+# pf9ctl
+
+### Status
 ![Go](https://github.com/roopakparikh/pf9ctl/workflows/Go/badge.svg)
 
-# pf9ctl
-CLI tool for Platform9 management. This is under heavy development, please use with care.
+### Purpose
+   CLI tool for Platform9 management. This is under heavy development, please use with care
+   
+### Requirements
 
-To disable the segment event set environment variable. PF9CTL_SEGMENT_EVENTS_DISABLE=1
+- CPUs: Minimum 2 CPUs needed on host
+- RAM: 12 GB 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,266 @@
...
+
+Global Flags:
+      --verbose   print verbose logs
+
+Use "pf9ctl config [command] --help" for more information about a command.
+```  
+
+- **Check Node**
+
+  This command will perform the prerequisite check before a node can be added to the cluster. It checks for the following:
```
anupbarve, line 84: s/check/checks
Could we add which checks are mandatory and which ones are optional ?

README.md
```diff
@@ -1,6 +1,266 @@
...
+✓ DiskCheck
+✓ MemoryCheck
+✓ PortCheck
+✓ Existing Kubernetes Cluster Check
+
+✓ Completed Pre-Requisite Checks successfully
+```
+- **prep-node**
+
+ This command prepares a node to be ready to be added to a cluster. It installs all the required pre-requisites on the nodes and installs the Platform9 host agent and other required software on the nodes.
```
anupbarve, line 155: Replace above text with:

This command onboards a node. It installs platform9 packages on the host. After completion of this command, the node is available to be managed on the Platform9 control plane.

anupbarve (changes requested):

README.md
ShaunakJoshi1407: Ok will add that

README.md
ShaunakJoshi1407: Ok will replace this text

anupbarve (approved):

PR #155, opened 2021-05-18
Moved swap off to check-node

devidask27:
Moved swap off functionality from prep-node to check-node

./pf9ctl prep-node -i 172.20.7.2 -u ubuntu -s /home/ubuntu/.ssh/mykey
✓ Loaded Config Successfully
✓ Removal of existing CLI
✓ Existing Platform9 Packages Check
✓ Required OS Packages Check
✓ SudoCheck
✓ CPUCheck
✓ DiskCheck
✓ MemoryCheck
✓ PortCheck
✓ Existing Kubernetes Cluster Check
✓ Disabling swap and removing swap in fstab

✓ Completed Pre-Requisite Checks successfully

------------------------------------------------------------
./pf9ctl prep-node -i 172.20.7.109 -u ubuntu -s /home/ubuntu/.ssh/mykey --disableSwapOff
✓ Loaded Config Successfully
✓ Removal of existing CLI
x Existing Platform9 Packages Check - Platform9 packages already exist. These must be uninstalled.
✓ Required OS Packages Check
✓ SudoCheck
✓ CPUCheck
✓ DiskCheck
✓ MemoryCheck
x PortCheck - Following port(s) should not be in use: 10250, 3306
x Existing Kubernetes Cluster Check - A Kubernetes cluster is already running on node

x Required pre-requisite check(s) failed.

### DIFF
--- a/pkg/platform/centos/centos.go
+++ b/pkg/platform/centos/centos.go
@@ -296,7 +296,7 @@ func (c *CentOS) installOSPackages(p string) error {
 func (c *CentOS) disableSwap() (bool, error) {
 	err := swapoff.SetupNode(c.exec)
 	if err != nil {
-		return false, errors.New("error occured while removing swap")
+		return false, errors.New("error occured while disabling swap")
 	} else {
 		return true, nil
 	}
--- a/pkg/platform/centos/centos.go
+++ b/pkg/platform/centos/centos.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/platform9/pf9ctl/pkg/cmdexec"
 	"github.com/platform9/pf9ctl/pkg/platform"
+	"github.com/platform9/pf9ctl/pkg/swapoff"
 	"github.com/platform9/pf9ctl/pkg/util"
 	"go.uber.org/zap"
 )
@@ -61,6 +62,11 @@ func (c *CentOS) Check() []platform.Check {
 
 	result, err = c.checkKubernetesCluster()
 	checks = append(checks, platform.Check{"Existing Kubernetes Cluster Check", true, result, err, fmt.Sprintf("%s", err)})
+
+	if !util.SwapOffDisabled {
+		result, err = c.disableSwap()
+		checks = append(checks, platform.Check{"Disabling swap and removing swap in fstab", true, result, err, fmt.Sprintf("%s", err)})
+	}
 
 	return checks
 }
@@ -268,7 +274,7 @@ func (c *CentOS) Version() (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("Couldn't read the OS configuration file os-release: %s", err.Error())
 	}
-	if match, _:= regexp.MatchString(`.*7\.[3-9]\.*|.*8\.[3]\.*`, string(out)); match {
+	if match, _ := regexp.MatchString(`.*7\.[3-9]\.*|.*8\.[3]\.*`, string(out)); match {
 		return "redhat", nil
 	}
 	return "", fmt.Errorf("Unable to determine OS type: %s", string(out))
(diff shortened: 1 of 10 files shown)


pkg/pmk/checkNode.go
```diff
@@ -85,6 +86,11 @@ func CheckNode(ctx Config, allClients Client) (CheckNodeResult, error) {
 	defer s.Stop()
 	s.Suffix = " Running pre-requisite checks and installing any missing OS packages"
 	checks := platform.Check()
+	if !SwapOffDisabled {
+		checks = disableSwap(os, allClients.Executor, checks)
+	} else {
+		zap.S().Debug("disableSwapOff is set, not disabling the swap")
+	}
```
anupbarve, line 93: Let us not convolute the logic with some checks here and some in platform module. 
It becomes very hard to maintain such code.
Could you please treat this check similar to other checks and make it a mandatory one ?

anupbarve (changes requested):

anupbarve:
Could you also test the failure case, where disable swapoff failure is manually induced and then check-node detects it.

pkg/platform/debian/debian.go
```diff
@@ -61,6 +62,10 @@ func (d *Debian) Check() []platform.Check {
 	result, err = d.checkKubernetesCluster()
 	checks = append(checks, platform.Check{"Existing Kubernetes Cluster Check", true, result, err, fmt.Sprintf("%s", err)})
 
+	if !util.SwapOffDisabled {
```
anupbarve, line 65: Can we move this check inside `disableSwap` swap function itself. 
This will make it self-sufficient.

pkg/platform/debian/debian.go
```diff
@@ -288,3 +293,12 @@ func (d *Debian) installOSPackages(p string) error {
 	_, err = d.exec.RunWithStdout("bash", "-c", fmt.Sprintf("apt install -qq -y %s", p))
 	return nil
 }
+
+func (d *Debian) disableSwap() (bool, error) {
+	err := swapoff.SetupNode(d.exec)
+	if err != nil {
+		return false, errors.New("error occured while removing swap")
```
anupbarve, line 300: s/removing/disabling

pkg/platform/centos/centos.go
```diff
@@ -286,3 +292,12 @@ func (c *CentOS) installOSPackages(p string) error {
 	_, err = c.exec.RunWithStdout("bash", "-c", fmt.Sprintf("yum -q -y install %s", p))
 	return nil
 }
+
+func (c *CentOS) disableSwap() (bool, error) {
+	err := swapoff.SetupNode(c.exec)
+	if err != nil {
+		return false, errors.New("error occured while removing swap")
```
anupbarve, line 299: s/removing/disabling

pkg/platform/centos/centos.go
```diff
@@ -62,6 +63,11 @@ func (c *CentOS) Check() []platform.Check {
 	result, err = c.checkKubernetesCluster()
 	checks = append(checks, platform.Check{"Existing Kubernetes Cluster Check", true, result, err, fmt.Sprintf("%s", err)})
 
+	if !util.SwapOffDisabled {
```
anupbarve, line 66: Can we move this to disableSwap function, to make it self-sufficient?

anupbarve (changes requested):
1.  Did you try to induce failure in swapoff disable function ?
2. Can we add as test case for this functionality in our existing test case infra ?

pkg/platform/debian/debian.go
devidask27: I think we cannot move. because whether to call disableSwap or not is depends on this flag.

pkg/platform/centos/centos.go
devidask27: I think we cannot move. because whether to call disableSwap or not is depends on this flag.

pkg/platform/debian/debian.go
anupbarve: I am proposing something like this:

`
disableSwap {
 if util.SwapOffDisabled
     return
logic for disableSwap
}
`

devidask27: yes we can do that. 
if util.SwapOffDisabled 
return false, nill

But check node will fail because this is mandatary check

anupbarve (approved):

PR #157, opened 2021-05-26
Modified the message to pass password during config set through (-p flag).

AnirudhPokala:
Config Set fails, if password contains special characters and passed through -p flag during config set. So, we should pass using single quotes. 

Screenshot 2021-05-26 at 3 29 14 PM

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -77,7 +77,7 @@ func configCmdCreateRun(cmd *cobra.Command, args []string) {
 		if err := validateUserCredentials(ctx, c); err != nil {
 
 			if SetConfigByParameters {
-				zap.S().Fatalf("Invalid credentials entered (Username/Password/Tenant)")
+				zap.S().Fatalf("Invalid credentials entered (Username/Password/Tenant), use 'single quotes' to pass password ")
 				credentialFlag = false
 			} else {
 				//Clearing the invalid config entered. So that it will ask for new information again.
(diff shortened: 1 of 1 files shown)


cmd/config.go
```diff
@@ -77,7 +77,7 @@ func configCmdCreateRun(cmd *cobra.Command, args []string) {
 		if err := validateUserCredentials(ctx, c); err != nil {
 
 			if SetConfigByParameters {
-				zap.S().Fatalf("Invalid credentials entered (Username/Password/Tenant)")
+				zap.S().Fatalf("Invalid credentials entered (Username/Password/Tenant), use 'single quotes' to pass password ")
```
nikhilbandi, line 80: Better one: "use 'single quotes' to enter password."
Please make this change at both places.

nikhilbandi (changes requested):
Please implement minor suggestion.

cmd/config.go
AnirudhPokala: But Nikhil, we are passing the password through flags right, so kept "to pass password", as we did the same in check-node, prep-node case.

AnirudhPokala: So should we change the same in check-node and prep-node case also?

nikhilbandi: OK then. To keep it consistent, let's keep 'pass password' here.
Although, 'enter' looks more appropriate. We will discuss & decide on that later.

nikhilbandi (approved):
Approving.

PR #163, opened 2021-06-13
Private/devidask/dpkg lock check

devidask27:
- Added pre-requisite check to check if there is lock on dpkg and apt.
- Added pre-requisite check to check if system is booted with systemd(PID=1)


anupbarve (changes requested):
Need to add tests for these checks.

pkg/platform/debian/debian.go
```diff
@@ -314,3 +323,36 @@ func (d *Debian) checkNoexecPermission() (bool, error) {
...
+}
+
+func (d *Debian) checkIfSystemIsBootedWithSystemd() (bool, error) {
+	_, err := d.exec.RunWithStdout("bash", "-c", "pidof systemd | grep '1$'")
+	if err != nil {
+		return false, errors.New("PID of systemd is not 1")
+	} else {
+		return true, nil
+	}
+}
```
anupbarve, line 358: Please remove this check from the current PR. 
We are not yet sure if it works.
Also remove the corresponding test cases. 
We will take it up as part of another PR.

devidask27: Okay

pkg/platform/centos/centos.go
```diff
@@ -313,3 +316,12 @@ func (c *CentOS) checkNoexecPermission() (bool, error) {
...
 }
+
+func (c *CentOS) checkIfSystemIsBootedWithSystemd() (bool, error) {
+	_, err := c.exec.RunWithStdout("bash", "-c", "pidof systemd | grep '1'")
+	if err != nil {
+		return false, errors.New("PID of systemd is not 1")
+	} else {
+		return true, nil
+	}
+}
```
anupbarve, line 327: Also need to remove this one

devidask27: Done

anupbarve (approved):

PR #165, opened 2021-06-16
Added validation for region during config set/config loading.  

AnirudhPokala:
* We bail out if user enters invalid region during config set multiple times or prompt if found invalid. 

Cases possible:
1) Only Region invalid during config set. -- "we will return an error saying region is invalid".
2) Multiple credentials invalid during config set. -- 
3) Only Region invalid during config loading  -- "we will return an error saying region is invalid".
4) Multiple credentials invalid during config loading.

If only region is invalid during config loading then the output is 
Screenshot 2021-06-16 at 12 51 52 PM

If region is invalid during config set. 
Screenshot 2021-06-16 at 12 54 28 PM

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -193,7 +193,7 @@ func validateUserCredentials(pmk.Config, pmk.Client) error {
 	endpointURL, err1 := pmk.FetchRegionFQDN(ctx, auth)
 	if endpointURL == "" || err1 != nil {
 		RegionInvalid = true
-		zap.S().Debug("Invalid Region entered")
+		zap.S().Debug("Invalid Region")
 		return errors.New("Invalid Region")
 	}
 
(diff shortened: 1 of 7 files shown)


cmd/config.go
```diff
@@ -77,13 +80,17 @@ func configCmdCreateRun(cmd *cobra.Command, args []string) {
 		if err := validateUserCredentials(ctx, c); err != nil {
 
 			if SetConfigByParameters {
-				zap.S().Fatalf("Invalid credentials entered (Username/Password/Tenant), use 'single quotes' to pass password ")
+				if RegionInvalid {
+					zap.S().Fatalf("Invalid credentials entered (Region)")
```
anupbarve, line 84: nit: Invalid Region instead of creds. Here and everywhere below.

cmd/config.go
```diff
@@ -171,15 +178,29 @@ func init() {
...
+
+	// To validate region.
+	endpointURL, err1 := pmk.FetchRegionFQDN(ctx, auth)
+	if endpointURL == "" || err1 != nil {
+		RegionInvalid = true
+		zap.S().Debug("Invalid Region entered")
+		return errors.New("Invalid Region")
+	}
+
+	return nil
```
anupbarve, line 200: I think that the correct way of doing this is :
- Fetch region FQDN (Handle success/failure)
- Use this region FQDN to fetch the keystone token. (Handle success/failure)

anupbarve: Can we also store the region fqdn in config.json ?

anupbarve (changes requested):

cmd/config.go
AnirudhPokala: Creating a new Jira to handle this as discussed.

anupbarve (approved):

PR #166, opened 2021-06-24
Added pre-check to check if system is booted with systemd as a init

devidask27:
systemctl command was failing because if system was not booted with systemd. So added pre-requisite check to check if system is booted with systemd.

### DIFF
--- a/pkg/platform/centos/centos.go
+++ b/pkg/platform/centos/centos.go
@@ -66,6 +66,9 @@ func (c *CentOS) Check() []platform.Check {
 	result, err = c.checkNoexecPermission()
 	checks = append(checks, platform.Check{"Check exec permission on /tmp", true, result, err, fmt.Sprintf("%s", err)})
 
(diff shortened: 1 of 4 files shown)


pkg/platform/centos/centos.go
```diff
@@ -66,6 +66,9 @@ func (c *CentOS) Check() []platform.Check {
 	result, err = c.checkNoexecPermission()
 	checks = append(checks, platform.Check{"Check exec permission on /tmp", true, result, err, fmt.Sprintf("%s", err)})
 
+	result, err = c.checkPIDofSystemd()
+	checks = append(checks, platform.Check{"Check if system booted with systemd", true, result, err, fmt.Sprintf("%s", err)})
```
anupbarve, line 70: nit:
s/system booted/system is booted
Also change in debian.go

nikhilbandi (approved):
Looks good except for the minor comment provided by Anup.

anupbarve (approved):

PR #168, opened 2021-06-25
added rhel os check

devidask27:
Added RHEL OS Support. Teated it on RHEL 7.4 and RHEL 7.5
Screenshot 2021-06-25 at 2 51 03 PM
Screenshot 2021-06-25 at 2 51 36 PM

### DIFF
--- a/pkg/platform/centos/centos.go
+++ b/pkg/platform/centos/centos.go
@@ -273,7 +273,7 @@ func (c *CentOS) Version() (string, error) {
 	out, err := c.exec.RunWithStdout(
 		"bash",
 		"-c",
-		"cat /etc/*release | grep 'CentOS Linux release' -m 1 | cut -f4 -d ' '")
+		"cat /etc/*release | grep 'Red Hat Enterprise Linux Server release' -m 1 | cut -f7 -d ' ' && cat /etc/*release | grep 'CentOS Linux release' -m 1 | cut -f4 -d ' '")
 	if err != nil {
 		return "", fmt.Errorf("Couldn't read the OS configuration file os-release: %s", err.Error())
 	}
--- a/pkg/pmk/node.go
+++ b/pkg/pmk/node.go
@@ -241,7 +241,7 @@ func ValidatePlatform(exec cmdexec.Executor) (string, error) {
 	}
 	var platform platform.Platform
 	switch {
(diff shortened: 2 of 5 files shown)


nikhilbandi:
@devidask27 , have you tested for negative test cases in case of check-node command?
Also, when we display error message in case of unsupported Operating Systems, we should include RHEL (with version numbers) as one of the supported OSes apart from Ubuntu & Centos. Has this been taken care of?

pkg/platform/centos/centos.go
```diff
@@ -273,7 +273,7 @@ func (c *CentOS) Version() (string, error) {
 	out, err := c.exec.RunWithStdout(
 		"bash",
 		"-c",
-		"cat /etc/*release | grep 'CentOS Linux release' -m 1 | cut -f4 -d ' '")
+		"cat /etc/*release | grep 'Red Hat Enterprise Linux Server release' -m 1 | cut -f7 -d ' ' && cat /etc/*release | grep 'CentOS Linux release' -m 1 | cut -f4 -d ' '")
```
anupbarve, line 276: Why are we adding rhel check in centos.go ?

pkg/pmk/node.go
```diff
@@ -241,7 +241,7 @@ func ValidatePlatform(exec cmdexec.Executor) (string, error) {
 	}
 	var platform platform.Platform
 	switch {
-	case strings.Contains(strData, util.Centos) || strings.Contains(strData, util.Redhat):
+	case strings.Contains(strData, util.Centos) || strings.Contains(strData, util.Redhat) || strings.Contains(strData, "Red Hat Enterprise Linux"):
```
anupbarve, line 244: Why is this check needed when we already have Redhat as 2nd check ?

devidask27: Removed

anupbarve (approved):

devidask27:
> @devidask27 , have you tested for negative test cases in case of check-node command?
> Also, when we display error message in case of unsupported Operating Systems, we should include RHEL (with version numbers) as one of the supported OSes apart from Ubuntu & Centos. Has this been taken care of?

Tested for negative test cases in case of check-node
Screenshot 2021-06-29 at 3 51 56 PM
Also checked for port. it is detecting if port is in use.

PR #173, opened 2021-07-07
Proxy

sahil-lakhwani:
Allow users to use a proxy for all the network requests made by the CLI and the host-agents to route through the specified proxy.

The proxy URL is stored in config and can be set by the existing command `pf9ctl config set`

### DIFF
--- a/pkg/pmk/config.go
+++ b/pkg/pmk/config.go
@@ -148,6 +148,12 @@ func ConfigCmdCreateRun() (Config, error) {
 		service, _ = reader.ReadString('\n')
 		Context.Tenant = strings.TrimSuffix(service, "\n")
 	}
+	var proxyURL string
+	if Context.ProxyURL == "" {
+		fmt.Print("Proxy URL: ")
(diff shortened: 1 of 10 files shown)


pkg/pmk/config.go
```diff
@@ -139,6 +148,12 @@ func ConfigCmdCreateRun() (Config, error) {
 		service, _ = reader.ReadString('\n')
 		Context.Tenant = strings.TrimSuffix(service, "\n")
 	}
+	var proxyURL string
+	if Context.ProxyURL == "" {
+		fmt.Print("Proxy URL: ")
```
anupbarve, line 153: Can we indicate that default proxy is None, just like we indicate that default region is [RegionOne] ?

anupbarve (approved):

pkg/pmk/config.go
sahil-lakhwani: Added `[None]` to the prompt text.

anupbarve (approved):

devidask27 (approved):

PR #176, opened 2021-07-13
Private/devidask/remove ntp installation

devidask27:


pkg/platform/debian/debian.go
```diff
@@ -356,3 +358,29 @@ func (d *Debian) checkPIDofSystemd() (bool, error) {
 		return true, nil
 	}
 }
+
+func (d *Debian) checkIfSystemdTimesyncdServiceRunning() {
```
nikhilbandi, line 362: I thought the logic was:
Check if timesyncd is running
  if not, check if ntp is running
    if not, start timesyncd, which id present by default in Ubuntu 16.04 + versions.

This should get rid of checking for any versions.

I am missing something here?

Also, please write this logic in the function in the form of comments.

pkg/platform/debian/debian.go
```diff
@@ -356,3 +358,29 @@ func (d *Debian) checkPIDofSystemd() (bool, error) {
...
+			} else {
+				zap.S().Debug("installing systemd-timesyncd")
+				d.installOSPackages("systemd-timesyncd")
+				zap.S().Debug("Starting systemd-timesyncd service")
+				if _, err := d.exec.RunWithStdout("bash", "-c", "systemctl start systemd-timesyncd"); err != nil {
+					zap.S().Fatalf("Failed to start systemd-timesyncd")
+				}
+			}
+		} else {
+			zap.S().Debug("NTP service is running. skiping the installation.")
```
nikhilbandi, line 381: skiping => Skipping

pkg/platform/debian/debian.go
```diff
@@ -356,3 +358,29 @@ func (d *Debian) checkPIDofSystemd() (bool, error) {
...
+		}
+	} else {
+		zap.S().Debug("systemd-timesyncd service is running. Skiping the installation.")
```
nikhilbandi, line 384: Skiping => Skipping

pkg/pmk/node.go
```diff
@@ -223,10 +223,10 @@ func installHostAgentCertless(ctx Config, regionURL string, auth keystone.Keysto
 	}
 
 	if IsRemoteExecutor {
-		cmd = fmt.Sprintf(`bash /tmp/installer.sh --no-proxy --skip-os-check --ntpd %s`, installOptions)
```
nikhilbandi, line 226: What's the significance of replacing option ntpd with no-ntp?

devidask27: it will skip ntp installation.

pkg/platform/debian/debian.go
devidask27: Okay i will do the change.

devidask27: Done

anupbarve (approved):

PR #183, opened 2021-08-02
Added pf9-packages name in check

devidask27:
Previously we were just checking for 'pf9-' . Now Full package name is added.

### DIFF
--- a/pkg/platform/centos/centos.go
+++ b/pkg/platform/centos/centos.go
@@ -106,13 +106,22 @@ func (c *CentOS) checkDocker() error {
 }
 
 func (c *CentOS) checkExistingInstallation() (bool, error) {
-
-	out, err := c.exec.RunWithStdout("bash", "-c", "yum list installed | { grep -i 'pf9-' || true; }")
-	if err != nil {
-		return false, nil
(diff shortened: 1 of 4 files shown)


pkg/platform/centos/centos.go
```diff
@@ -106,13 +106,22 @@ func (c *CentOS) checkDocker() error {
 }
 
 func (c *CentOS) checkExistingInstallation() (bool, error) {
-
-	out, err := c.exec.RunWithStdout("bash", "-c", "yum list installed | { grep -i 'pf9-' || true; }")
-	if err != nil {
-		return false, nil
+	var pf9Packages = []string{"pf9-hostagent", "pf9-comms", "pf9-kube", "pf9-muster"}
```
nikhilbandi, line 109: If we are hardcoding names of pf9 packages, would it make sense to have in the config file?

nikhilbandi (changes requested):
Please look at the minor comment.

pkg/platform/centos/centos.go
devidask27: Do you mean constant file.

nikhilbandi (approved):
Changes look good.

PR #186, opened 2021-08-05
checking if firewalld service is running

devidask27:
Bailing out if firewalld service is running on centos
Screenshot 2021-08-05 at 3 46 26 PM

### DIFF
--- a/pkg/platform/centos/centos.go
+++ b/pkg/platform/centos/centos.go
@@ -339,10 +339,10 @@ func (c *CentOS) checkPIDofSystemd() (bool, error) {
 }
 
 func (c *CentOS) checkFirewalldIsRunning() (bool, error) {
-	_, err := c.exec.RunWithStdout("bash", "-c", "systemctl status firewalld | grep 'active (running)'")
+	_, err := c.exec.RunWithStdout("bash", "-c", "systemctl is-active firewalld | grep 'inactive'")
 	if err != nil {
-		return true, nil
-	} else {
 		return false, errors.New("firewalld service is running")
+	} else {
+		return true, nil
 	}
 }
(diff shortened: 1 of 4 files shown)


pkg/platform/centos/centos.go
```diff
@@ -334,3 +337,12 @@ func (c *CentOS) checkPIDofSystemd() (bool, error) {
 		return true, nil
 	}
 }
+
+func (c *CentOS) checkFirewalldIsRunning() (bool, error) {
+	_, err := c.exec.RunWithStdout("bash", "-c", "systemctl status firewalld | grep 'active (running)'")
```
sahil-lakhwani, line 342: do we need `bash -c` ?

devidask27: Yes

pkg/platform/centos/centos.go
```diff
@@ -69,6 +69,9 @@ func (c *CentOS) Check() []platform.Check {
 	result, err = c.checkPIDofSystemd()
 	checks = append(checks, platform.Check{"Check if system is booted with systemd", true, result, err, fmt.Sprintf("%s", err)})
 
+	result, err = c.checkFirewalldIsRunning()
+	checks = append(checks, platform.Check{"Check if firewalld service is running", true, result, err, fmt.Sprintf("%s", err)})
```
anupbarve, line 73: This check needs to be optional.

devidask27: Okay

pkg/platform/centos/centos.go
anupbarve, line 342: You could check if the following command is useful
`systemctl is-active firewalld`

anupbarve (commented):
I was able to install firewalld on my ubuntu machine as well.
We need to cover debian based checks as well.

PR #187, opened 2021-08-11
Added check to check chronyd check

devidask27:


pkg/platform/debian/debian.go
```diff
@@ -371,20 +371,37 @@ func (d *Debian) checkPIDofSystemd() (bool, error) {
...
+							return true, nil
+						}
+					} else {
+						zap.S().Debug("Started chronyd service")
+						return true, nil
+					}
+				} else {
+					zap.S().Debug("Started systemd-timesyncd service")
+					return true, nil
+				}
```
anupbarve, line 398: This style of code is not readable nor maintainable. 
Split the logic up in small chunks as we had discussed earlier.
Also, I would make this check optional.

pkg/platform/debian/debian.go
```diff
@@ -371,29 +371,138 @@ func (d *Debian) checkPIDofSystemd() (bool, error) {
...
+		zap.S().Debug("Could not find os version")
+	}
+	return out
+}
+
+func (d *Debian) IsPresent(service string) error {
+	zap.S().Debugf("checking if %s is present", service)
+	var cmd string
+	version := d.getVersion()
+	if strings.Contains(string(version), "16") {
```
anupbarve, line 441: I don't think this is correct way of confirming versions.
You should filter major version first and then compare.
Also move the hardcoded numbers in constants.go.
The comment applies here and all other relevant places.

pkg/platform/debian/debian.go
```diff
@@ -371,29 +371,138 @@ func (d *Debian) checkPIDofSystemd() (bool, error) {
...
+	return err
+}
+
+func (d *Debian) getVersion() string {
+	out, err := d.exec.RunWithStdout("bash", "-c", "cat /etc/*os-release | grep -i pretty_name | cut -d ' ' -f 2")
+	if err != nil {
+		zap.S().Debug("Could not find os version")
+	}
+	return out
+}
```
anupbarve, line 435: Instead of implementing this here, we should add this functionality in a similar way like https://github.com/platform9/pf9ctl/blob/b3de43deeed1b05afa164ca64e44e7e83134293c/pkg/platform/centos/centos.go#L284
Can be taken in another PR, not a blocker for this one.

pkg/platform/debian/debian.go
```diff
@@ -371,29 +371,139 @@ func (d *Debian) checkPIDofSystemd() (bool, error) {
...
+		zap.S().Debugf("%s is started", service)
+		return nil
+	}
+}
+
+func (d *Debian) DownloadAndInstallTimesyncPkg() error {
+	zap.S().Debug("timesync package not found installing timesync package")
+	version := d.getVersion()
+	var err error
+	if strings.Contains(string(version), "20") {
```
anupbarve, line 496: your major version check also needs to be incorporated here. 
Also check for any other such missing places in this PR.

anupbarve (approved):

PR #188, opened 2021-08-23
downloading installer into pf9 dir instead of /tmp

devidask27:


nikhilbandi (commented):
Hope you have tested deploying in directories other than the default /home/ubuntu one.
Looks fine in general.

sahil-lakhwani:
@devidask27 Should the `pf9` directory be under HOME? In the current change it seems it will exist from where the CLI is run. Let me know if I'm mistaken here

devidask27:
> @devidask27 Should the `pf9` directory be under HOME? In the current change it seems it will exist from where the CLI is run. Let me know if I'm mistaken here

@sahil-lakhwani Yes it is under HOME. When we we download binary pf9 folder is created in home directory.

pkg/pmk/node.go
```diff
@@ -208,7 +208,7 @@ func installHostAgentCertless(ctx Config, regionURL string, auth keystone.Keysto
 	if ctx.AllowInsecure {
 		insecureDownload = "-k"
 	}
-	cmd := fmt.Sprintf(`curl %s --silent --show-error  %s -o  /tmp/installer.sh`, insecureDownload, url)
+	cmd := fmt.Sprintf(`curl %s --silent --show-error  %s -o  pf9/installer.sh`, insecureDownload, url)
```
sahil-lakhwani, line 211: @devidask27 Should this be `~/pf9/installer.sh` ?
You can of course get the home directory using `user.HomeDir`
Same for other places in this file

anupbarve: agree with @sahil-lakhwani

pkg/pmk/node.go
```diff
@@ -235,6 +238,21 @@ func installHostAgentCertless(ctx Config, regionURL string, auth keystone.Keysto
...
+		zap.S().Debug("error removing temporary directory")
+	}
+
+	zap.S().Debug("Removing installer script")
+	removeInstallerCmd := fmt.Sprintf("rm -rf %s/pf9/installer.sh", homeDir)
+	_, err1 = exec.RunWithStdout("bash", "-c", removeInstallerCmd)
+	if err1 != nil {
+		zap.S().Debug("error removing installer script")
+	}
+
```
anupbarve, line 255: Fold this into a function.

pkg/pmk/node.go
```diff
@@ -307,27 +325,49 @@ func installHostAgentLegacy(ctx Config, regionURL string, auth keystone.Keystone
...
+	if err1 != nil {
+		zap.S().Debug("error removing temporary directory")
+	}
+
+	zap.S().Debug("Removing installer script")
+	removeInstallerCmd := fmt.Sprintf("rm -rf %s/pf9/installer.sh", homeDir)
+	_, err1 = exec.RunWithStdout("bash", "-c", removeInstallerCmd)
+	if err1 != nil {
+		zap.S().Debug("error removing installer script")
 	}
```
anupbarve, line 369: Reuse the function created as a part of above comment.

anupbarve (changes requested):

anupbarve (approved):

anupbarve (approved):

PR #189, opened 2021-08-24
Removed the user password as part of CLI logs, and removed config.json as a part of supportbundle

AnirudhPokala:
To handle security issues with password being printed in logs, and stored as part of supportbundle,
* Now CLI logs wont print password as part of installer command. 
Screenshot 2021-08-24 at 6 07 21 PM

* The config.json is removed from supportbundle as part of ~/pf9/db
Screenshot 2021-08-24 at 6 40 04 PM

### DIFF
--- a/pkg/cmdexec/executor.go
+++ b/pkg/cmdexec/executor.go
@@ -4,6 +4,7 @@ package cmdexec
 import (
 	"fmt"
 	"os/exec"
+	"strings"
 
 	"github.com/platform9/pf9ctl/pkg/ssh"
 	"go.uber.org/zap"
@@ -53,7 +54,16 @@ func (c LocalExecutor) RunWithStdout(name string, args ...string) (string, error
 	if exitError, ok := err.(*exec.ExitError); ok {
 		stderr = string(exitError.Stderr)
 	}
-	zap.S().Debug("Ran command ", "sudo", args)
+
+	// To append args to a single command
+	command := ""
+	for _, arg := range args {
+		command = fmt.Sprintf("%s \"%s\"", command, arg)
+	}
+	// Remove user password if command has passwordflag
+	command = PasswordRemover(command)
+	zap.S().Debug("Ran command sudo", command)
+
 	zap.S().Debug("stdout:", string(byt), "stderr:", stderr)
 	return string(byt), err
 }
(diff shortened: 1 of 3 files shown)


pkg/cmdexec/executor.go
```diff
@@ -53,7 +54,16 @@ func (c LocalExecutor) RunWithStdout(name string, args ...string) (string, error
...
 		stderr = string(exitError.Stderr)
 	}
-	zap.S().Debug("Ran command ", "sudo", args)
+
+	// To append args to a single command
+	command := ""
+	for _, arg := range args {
+		command = fmt.Sprintf("%s \"%s\"", command, arg)
+	}
+	// Remove user password if command has passwordflag
```
nikhilbandi, line 63: nit: //Avoid password from getting logged, if the command contains password flag

pkg/cmdexec/executor.go
```diff
@@ -82,7 +92,11 @@ func (r *RemoteExecutor) RunWithStdout(name string, args ...string) (string, err
 	stdout, stderr, err := r.Client.RunCommand(cmd)
 	// To fetch the stderr after executing command.
 	StdErrSudoPassword = string(stderr)
-	zap.S().Debug("Running command ", cmd, "stdout:", string(stdout), "stderr:", string(stderr))
+
+	// Remove user password if command has passwordflag
```
nikhilbandi, line 96: nit: //Avoid password from getting logged, if the command contains password flag

pkg/supportBundle/supportBundle.go
```diff
@@ -215,7 +215,7 @@ func GenSupportBundle(exec cmdexec.Executor, timestamp time.Time) (string, error
 
 	} else {
 		// Generation of supportBundle in local host case.
-		_, errbundle := exec.RunWithStdout("bash", "-c", fmt.Sprintf("tar czf %s --directory=%s pf9 %s %s",
+		_, errbundle := exec.RunWithStdout("bash", "-c", fmt.Sprintf("tar czf %s --directory=%s pf9/log %s %s",
```
nikhilbandi, line 218: Can we replace pf9/log with the constant util.Pf9LogDir ?

nikhilbandi (commented):
Added a few minor comments. Otherwise looks good.
Hope you have tested for hypothetical cases like the flag --pwd etc.

anupbarve (dismissed):
As a quick fix, I am okay with this change.
But I am not comfortable with inserting command specific logic in the infrastructure function.

Ideally, what needs to be logged should also be passed by the caller. But I suppose that will need changes to all the instances of the code.

pkg/supportBundle/supportBundle.go
AnirudhPokala: util.Pf9LogDir = /home/ubuntu/pf9/log, where as we need pf9/log. I will create another variable for it.

AnirudhPokala: Added a constant in utils for it.

nikhilbandi (approved):
Looks good.

PR #195, opened 2021-09-16
Cloud provider checks

JasminAnt:
Added cloud provider checks for Amazon, Azure and Google using their Golang SDK's.


pkg/pmk/cloud_providers.go
```diff
@@ -0,0 +1,469 @@
...
+	ctx := context.Background()
+
+	//the user sends the path to json file
+	iamService, err := iamGoogle.NewService(ctx, option.WithCredentialsFile(path))
+	if err != nil {
+		return fmt.Errorf(err.Error())
+	}
+
+	//the four roles needed
+	names := []string{
```
nikhilbandi, line 36: Instead of hardcoding here, can we it stored in something like constants.go file?

pkg/pmk/cloud_providers.go
```diff
@@ -0,0 +1,469 @@
...
+}
+
+//////////////////////////////////////////////
+
+func CheckAmazonPovider(username, awsID, awsSecret string) error {
+
+	//creates a new session using static credentials that the user passes and "" stands for token that is of no use in this situation
+	svc := iamAws.New(session.New(
+		&aws.Config{
+			Region:      aws.String("us-east-1"),
```
nikhilbandi, line 65: Why do we need to hardcode the region as "us-east-1" here?

pkg/pmk/cloud_providers.go
```diff
@@ -0,0 +1,469 @@
...
+	resultUser, errUser := svc.GetUser(inputUser)
+	if errUser != nil {
+		return fmt.Errorf("Amazon user error " + errUser.Error())
+	}
+	arn := resultUser.User.Arn
+
+	//checkpermission function takes the arn, svc and an array of permissions needed and checks if the user has all of them
+	//this can be easily upgraded by just copy pasting one checkPermission block of code and changing the permission array
+	if checkPermissions(arn, svc, []string{
+		"elasticloadbalancing:AddTags",
```
nikhilbandi, line 84: Again, no need to hardcode here.

pkg/pmk/cloud_providers.go
```diff
@@ -0,0 +1,469 @@
...
+	zoneSvc := ec2.New(zoneSess)
+
+	resultAvalZones, err := zoneSvc.DescribeAvailabilityZones(nil)
+	if err != nil {
+		return fmt.Errorf("Availability Zones error" + err.Error())
+	}
+
+	//checks if the user less than 2 availability zones
+	if len(resultAvalZones.AvailabilityZones) < 2 {
+		return fmt.Errorf("Availability Zones error")
```
nikhilbandi, line 128: Can we print more intuitive error message here, e.g. "Minimum 2 availability zones required."

pkg/pmk/cloud_providers.go
JasminAnt: Yes, I will move all constants to the constants.go file.

pkg/pmk/cloud_providers.go
JasminAnt: I remember reading it is something like a default region, but I can add so that users can also pass the region and if they dont  it will use us-east-1.

pkg/pmk/cloud_providers.go
JasminAnt: Sure, it will make more sense for the users.

pkg/pmk/cloud_providers.go
```diff
@@ -0,0 +1,469 @@
...
+	for i := range actions {
+		actionNames = append(actionNames, &actions[i])
+	}
+	return actionNames
+
+}
+
+//////////////////////////////////////////
+
+func CheckAzureProvider(tenantID, appID, subID, secretKey string) error {
```
nikhilbandi, line 345: If there are different functions and checks for different cloud providers, can we have separate files for each cloud provider?

pkg/pmk/cloud_providers.go
```diff
@@ -0,0 +1,469 @@
...
+	if checkArray(result.EvaluationResults) == false {
+		fmt.Println("Access Error")
+		return false
+	}
+
+	return true
+
+}
+
+func checkArray(results []*iamAws.EvaluationResult) bool {
```
nikhilbandi, line 319: Can the function name be more intuitive e.g., checkAllowed / checkIfAllowed?

pkg/pmk/cloud_providers.go
JasminAnt: I asked if I should make one or three files and was told to make only one file. I could split them into three files if needed.

pkg/pmk/cloud_providers.go
JasminAnt: Yes, will rename it.

anupbarve (commented):
Few generic comments before I dig deep into the code:

- Seems like you have added pf9ctl binary to github repo, which is why it is showing up as a diff. This needs to be removed.
- Could you please upload the testing snippets in the PR description. For example: https://github.com/platform9/pf9ctl/pull/191
- We would need to figure out a way to unit test the code.

pkg/pmk/cloud_providers.go
anupbarve: Let us have users pass this to the CLI.

pkg/pmk/cloud_providers.go
```diff
@@ -0,0 +1,469 @@
...
+
+	}
+
+	//if the result.Value lengt is 0 then the loop never happens and that means the principal was never found
+	fmt.Println("Principal not found")
+	return nil
+
+}
+
+func returnEnv(appID, tenantID, subID, secretKey string) {
```
nikhilbandi, line 425: Function name suggests some values are being returned, but implementation shows some env variables are being set.

nikhilbandi (changes requested):
Added a few comments. Please check.

pkg/pmk/cloud_providers.go
JasminAnt: Bag logic in naming the method from my side, renamed to setEnvs

anupbarve (changes requested):
Recording the comment that we discussed in the standup.
We should cover all the checks instead of exiting after first failure.

pkg/pmk/checkAzureProvider.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func CheckRoleAssignment(result assignment.RoleAssignmentListResult, subID string) bool {
+
+	//Iterates through the list
+	for _, s := range *result.Value {
+
+		//The contributor role has a static ID so it checks the RoleDefiniitonID of the service principal by comparing it to
+		//a string where only the subscriptionID is different
+		if *s.Properties.RoleDefinitionID == "/subscriptions/"+subID+"/providers/Microsoft.Authorization/roleDefinitions/b24988ac-6180-42a0-ab88-20f7382dd24c" {
```
anupbarve, line 102: Move the static ID in constants file.

pkg/pmk/checkAzureProvider.go
```diff
@@ -0,0 +1,156 @@
...
+	oldAppID := os.Getenv("AZURE_CLIENT_ID")
+	oldTenantID := os.Getenv("AZURE_TENANT_ID")
+	oldSecretKey := os.Getenv("AZURE_CLIENT_SECRET")
+	oldSubID := os.Getenv("AZURE_SUBSCRIPTION_ID")
+
+	//Sets the new data as environment so that azure sdk can use it correctly
+	os.Setenv("AZURE_CLIENT_ID", appID)
+	os.Setenv("AZURE_CLIENT_SECRET", secretKey)
+	os.Setenv("AZURE_TENANT_ID", tenantID)
+	os.Setenv("AZURE_SUBSCRIPTION_ID", subID)
```
anupbarve, line 31: Since you have already defined setEnvs method, you can use it here as well.
Also do we also need to do this for other cloud providers as well ?

pkg/pmk/config.go
```diff
@@ -117,6 +136,100 @@ func LoadConfig(loc string) (Config, error) {
 
 var Context Config
 
+func ConfigCmdCreateAmazonRun() (Config, error) {
```
anupbarve, line 139: We would also have to add an ability to run this by accepting the creds via command line parameters.
This would be helpful for automation. Prompts break automation.

anupbarve (changes requested):
Added few comments.
Also please update README.md with the necessary documentation for the newly introduced  commands.

pkg/pmk/checkAzureProvider.go
JasminAnt: Other cloud providers can take variables when creating clients while Azure can either use env, read from the a file or from azures cli (which reads it from the env).

pkg/pmk/config.go
JasminAnt: I could make it so that if no arguments are sent it reads from the console, otherwise it reads the arguments?

JasminAnt:
Example of Amazon EvaluationResult (elasticloadbalancing:ApplySecurityGroupsToLoadBalancer permission is granted but backup:CopyTarget permission is not granted so it is not giving "allowed" in EvalDecision)
image

cmd/cloudProviders.go
```diff
@@ -10,51 +10,89 @@ import (
...
+		}
+
+		if len(args) == 4 {
+			azureTenantID = args[0]
+			azureAppID = args[1]
+			azureSubID = args[2]
+			azureSecretKey = args[3]
+			loadConfig = false
+		} else {
+			loadConfig = true
```
anupbarve, line 93: Could we implement these parameters using cobra infra like this : https://github.com/platform9/pf9ctl/blob/b3de43deeed1b05afa164ca64e44e7e83134293c/pkg/platform/centos/centos.go#L284
This would mean that we take away order dependency for passing parameters.
Same comment for all cloud providers.

JasminAnt: Done 
image

anupbarve (approved):

PR #197, opened 2021-09-30
Added CentOS 8.4 support

devidask27:


pkg/pmk/node.go
```diff
@@ -219,7 +219,7 @@ func installHostAgentCertless(ctx Config, regionURL string, auth keystone.Keysto
 	var installOptions string
 
 	//Pass keystone token if MFA token is provided
-	if (ctx.MfaToken != "") {
+	if ctx.MfaToken != "" {
```
anupbarve, line 222: Can you please decouple this MFA change and add it in separate PR?
We are thinking of releasing a new binary without MFA support since it's corresponding change in pf9-du is still not released (will be done in 5.4).

PR #200, opened 2021-10-04
improved os version check for ubuntu

devidask27:
Currently we are comparing version with 16, 18, 20. which will cause check-node. and proceed for version like 20.10 which we dont support.

### DIFF
--- a/pkg/platform/debian/debian.go
+++ b/pkg/platform/debian/debian.go
@@ -299,11 +299,11 @@ func (d *Debian) Version() (string, error) {
 		"bash",
 		"-c",
 		"cat /etc/*os-release | grep -i pretty_name | cut -d ' ' -f 2")
-	mjrVersion := out[0:2]
+	mjrVersion := out[0:5]
 	if err != nil {
 		return "", fmt.Errorf("Couldn't read the OS configuration file os-release: %s", err.Error())
 	}
-	if strings.Contains(string(mjrVersion), "16") || strings.Contains(string(mjrVersion), "18") || strings.Contains(string(mjrVersion), "20") {
+	if strings.Contains(string(mjrVersion), "16.04") || strings.Contains(string(mjrVersion), "18.04") || strings.Contains(string(mjrVersion), "20.04") {
 		return "debian", nil
 	}
 	return "", fmt.Errorf("Unable to determine OS type: %s", string(out))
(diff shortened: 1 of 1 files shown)


pkg/platform/debian/debian.go
```diff
@@ -439,8 +439,8 @@ func (d *Debian) IsPresent(service string) error {
 	zap.S().Debugf("checking if %s is present", service)
 	var cmd string
 	version := d.getVersion()
-	mjrVersion := version[0:2]
-	if strings.Contains(string(mjrVersion), "16") {
+	mjrVersion := version[0:5]
+	if strings.Contains(string(mjrVersion), "16.04") {
```
anupbarve, line 443: Instead of duplicating the code, could we have `getVersion` to return major, minor and patch versions and make decisions based on those?

pkg/platform/debian/debian.go
```diff
@@ -427,20 +426,27 @@ func (d *Debian) checkIfAnyTimeSyncServiceIsRunning() error {
...
+	minor, err := d.exec.RunWithStdout("bash", "-c", "cat /etc/*os-release | grep -i pretty_name | cut -d ' ' -f 2 | cut -d.  -f 2")
+	if err != nil {
+		zap.S().Debug("Error occured while parsing minor os version")
+	}
+	patch, err := d.exec.RunWithStdout("bash", "-c", "cat /etc/*os-release | grep -i pretty_name | cut -d ' ' -f 2 | cut -d.  -f 3")
+	if err != nil {
+		zap.S().Debug("Error occured while parsing patch os version")
+	}
+	return major, minor, patch
 }
```
anupbarve, line 443: The version needs to be fetched only once in memory and parsed.
No need to run the same command for 3 times.

pkg/platform/debian/debian.go
```diff
@@ -299,11 +299,11 @@ func (d *Debian) Version() (string, error) {
 		"bash",
 		"-c",
 		"cat /etc/*os-release | grep -i pretty_name | cut -d ' ' -f 2")
-	mjrVersion := out[0:2]
+	mjrVersion := out[0:5]
```
anupbarve, line 302: can't we use getVersion() function here as well ?

anupbarve (approved):
lgtm

PR #202, opened 2021-10-05
removed CentOS 8 support

devidask27:


anupbarve (commented):
Please remove support for ubuntu16.04 as well as per following jira:
https://platform9.atlassian.net/browse/FT-223

anupbarve (approved):

PR #204, opened 2021-10-05
Fixed some command exit codes

JasminAnt:
Fixed a couple of commands not changing the exit code when they are stopped or when they finish unsuccessfully.

Prep node:
ExitCodeEg1

Check node:
ExitCodeEg2

Other commands seam to send the correct exit code, only the check-node and prep-node commands had an issue.

### DIFF
--- a/cmd/prepNode.go
+++ b/cmd/prepNode.go
@@ -120,21 +120,19 @@ func prepNodeRun(cmd *cobra.Command, args []string) {
 	}
 
 	if result == pmk.RequiredFail {
-		fmt.Println(color.Red("x ") + "Required pre-requisite check(s) failed.")
-		return
+		zap.S().Fatalf(color.Red("x ") + "Required pre-requisite check(s) failed.")
 	} else if !skipChecks {
 		if result == pmk.OptionalFail {
 			fmt.Print("\nOptional pre-requisite check(s) failed. Do you want to continue? (y/n) ")
 			reader := bufio.NewReader(os.Stdin)
 			char, _, _ := reader.ReadRune()
 			if char != 'y' {
-				return
+				os.Exit(0)
 			}
 		}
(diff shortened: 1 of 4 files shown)


cmd/root.go
```diff
@@ -36,8 +36,8 @@ var rootCmd = &cobra.Command{
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
 	if err := initializeBaseDirs(); err != nil {
-		fmt.Printf("Base directory initialization failed: %s\n", err.Error())
-		os.Exit(1)
+		//Fatalf does the same as Printf and os.Exit
```
sahil-lakhwani, line 39:
```suggestion
		// Fatalf does the same as Printf and os.Exit
```

cmd/prepNode.go
```diff
@@ -120,21 +120,19 @@ func prepNodeRun(cmd *cobra.Command, args []string) {
...
-		return
+		zap.S().Fatalf(color.Red("x ") + "Required pre-requisite check(s) failed.")
 	} else if !skipChecks {
 		if result == pmk.OptionalFail {
 			fmt.Print("\nOptional pre-requisite check(s) failed. Do you want to continue? (y/n) ")
 			reader := bufio.NewReader(os.Stdin)
 			char, _, _ := reader.ReadRune()
 			if char != 'y' {
-				return
+				os.Exit(1)
```
sahil-lakhwani, line 130: I am thinking if the exit code should be 1 here? The user chooses to not proceed. 
Although, the checks fail, the prep-node command doesn't.

JasminAnt: Yeah, my bad. Didn't notice it was an optional check. Thank you for noticing that.

anupbarve (approved):

PR #206, opened 2021-10-07
Fixed spinner filling console with symbols

JasminDudicBloomteq:
Spinner was filling up the console with some symbols. The problem was the spinner wasn't updated to the latest version which had fixed the bug. https://github.com/briandowns/spinner/pull/115 

Before updating, running on linux screen:
SpinnerError1

After updating, running on linux screen:
SpinnerError2

Note: I had added the spinner to the version command so that is why I'm running it. There is no spinner on the version command, it was just for testing.


anupbarve (approved):
Please run basic operations like check-node and prep-node to confirm that things are working as expected (specifically the console prints).
You can go ahead with the merge once done with it.

JasminAnt:
Spinner1
Spinner2

Tested on linux screen. These are the screenshots of the spinner in action.

PR #207, opened 2021-10-08
Added command to upgrade the cli verison.

JasminDudicBloomteq:
Note: Added changelog string to constants as a way to let the users know whats new in the newest version. Its just a string that will be written when the version command is called.

Test1: Testing input with y/n and anything else. If y/n are passed exit code is 0, if anything else is passed exit code is 1.
image
image

Test2: Running any command will call for a version check. Since AWS SDK was used for comparing files this doesn't take long. If the users internet is slow it might take a couple of seconds otherwise it should be instantaneous.
Upgrade2

Note: For automation purposes there is a --skipCheck flag ( or -c) that can be passed to skip the check and prompt.
image


sahil-lakhwani:
hey @JasminAnt , I think all the file operations will be local to where the CLI running - so we can prefer using Go SDKs for those executions instead of bash commands. WDYT?

pkg/util/constants.go
```diff
@@ -84,5 +84,6 @@ func init() {
 
 //This is the current version of cli.
 const (
-	Version string = "pf9ctl version: v1.8"
+	Version     string = "v1.8"
+	VersionPath string = "https://jasmind.s3-us-west-2.amazonaws.com/version"
```
anupbarve, line 88: Could we get away with storing another file for version ? This adds an additional responsibility of releasing the version string to the S3 bucket.

Could you please explore a possibility of fetching md5sum of the binary using AWS SDK ?
Can we simply check md5sum of local binary vs remote binary.

PR #209, opened 2021-10-13
Changed how installing works

JasminDudicBloomteq:
Made it so that the new installation is saved under a different name until every check passes and then it renames it. This is done so that the old version is not overwritten during the installation so that if the installation fails it will keep the old version.

Test1: Old installation was canceled and the old version was deleted
image

Test2: New installation was canceled but the old version was still there
image

### DIFF
--- a/pf9ctl_setup
+++ b/pf9ctl_setup
@@ -68,8 +68,11 @@ refresh_symlink() {
 }
 
 check_installation() {
-    if ! (${cli_exec} --help >> ${log_file} 2>&1); then
+    # Checking if the new "pf9ctl_new" installation is completed since the "pf9ctl" isn't removed untill the end of the installation process
+    if ! (${cli_exec}_new --help >> ${log_file} 2>&1); then
 	assert "Installation of Platform9 CLI Failed"; fi
+    # As this is the final check and everything is alright that means we can now rename the "pf9ctl_new" to "pf9ctl"
+    sudo mv ${cli_exec}_new ${cli_exec}
 }
 
 download_cli_binary() {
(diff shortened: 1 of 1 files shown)


pf9ctl_setup
```diff
@@ -62,24 +62,25 @@ refresh_symlink() {
...
 download_cli_binary() {
 	echo "Note: SUDO access required to run Platform9 CLI."
     echo "      You might be prompted for your SUDO password."
 	echo ""
 	echo "Downloading Platform9 CLI binary..."
-	sudo rm ${cli_exec} 2> /dev/null
-	cd ${pf9_bin} > /dev/null && sudo curl -s -O ${cli_path} -o ${log_file} 2>&1 && cd - > /dev/null
-	sudo chmod 555 ${cli_exec}
+	sudo rm ${cli_exec}_new 2> /dev/null
+	cd ${pf9_bin} > /dev/null && sudo curl -s -o pf9ctl_new "${cli_path}" -o ${log_file} 2>&1 && cd - > /dev/null
```
nikhilbandi, line 82: `curl -s -o pf9ctl_new`
Shouldn't the option be '-O' here, which depicts file name?
Think '-o' option refers to logging the output.

Also, can't we mention the file name here as '${cli_exec}_new' similar to what mentioned at other places? Is there any specific reason behind mentioning it as 'pf9ctl_new'?

pf9ctl_setup
```diff
@@ -62,24 +62,25 @@ refresh_symlink() {
...
     if ! (sudo ln -s ${cli_exec} /usr/bin/pf9ctl >> ${log_file} 2>&1); then
-	    assert "Failed to create Platform9 CLI symlink in /usr/bin"; fi
+	    assert "Failed to create Platform9 CLI symlink in /usr/bin"; fi 
 }
 
 check_installation() {
-    if ! (${cli_exec} --help >> ${log_file} 2>&1); then
+    if ! (${cli_exec}_new --help >> ${log_file} 2>&1); then
 	assert "Installation of Platform9 CLI Failed"; fi
+    sudo mv ${cli_exec}_new ${cli_exec}
```
nikhilbandi, line 73: Again, a comment here stating why are we renaming it to original CLI name would be great.

pf9ctl_setup
```diff
@@ -68,18 +68,22 @@ refresh_symlink() {
...
 
 download_cli_binary() {
 	echo "Note: SUDO access required to run Platform9 CLI."
     echo "      You might be prompted for your SUDO password."
 	echo ""
 	echo "Downloading Platform9 CLI binary..."
-	sudo rm ${cli_exec} 2> /dev/null
-	cd ${pf9_bin} > /dev/null && sudo curl -s -O ${cli_path} -o ${log_file} 2>&1 && cd - > /dev/null
-	sudo chmod 555 ${cli_exec}
+	sudo rm ${cli_exec}_new 2> /dev/null
```
sahil-lakhwani, line 83: Do we need this line here? Because there ${cli_exec}_new will never exist before and after this script.

PR #210, opened 2021-10-19
Added commands ot authorize, deauthorize, detach node. Added delete cluster command.

JasminAnt:
Delete cluster command:

The user must pass either a name or the uuid of the cluster. Passing both takes the uuid into consideration. Passing neither prompts an error.
image
If the user passes a non existing cluster name:
image
If the user passes a non existing uuid:
image

Deauthorize node command:

If the node was deauthoriuzed successfully:
image
If the node is not authorized the user will get an error:
image

Authorize command:

If the node is deauthorized and it is successfully authorized:
image
If the node is already authorized or in the process of being authorized:
image

Detach node:

When the node is detached successfully:
image

If the node is not attached to a cluster it will note that it starts the detaching process but will not say what node was detached. This is since the API does not return an error if the node is not attached to anything.

Attaching a detached node back to its cluster:
image

Decommission node:

When the node is decommissioned successfully:
image
...
image

If there is no host agent installed:
image


anupbarve (approved):
lgtm

PR #220, opened 2021-11-01
Automation

sahil-lakhwani:
Thia PR is to make pf9ctl be allowed to run without any user prompt
Two major additions are:
- `--no-prompt` flag -  Adding this flag will disable all user prompts. Any errors (missing flags, etc) are reported immediately and the cLI exits with non 0 code
 - `--mfa` - This is for commands like check-node & prep-node to allow users to push current MFA token

Testing done for these cases:
 - check-node
       - with `--no-prompt` and config set & not set
       - without `--no-prompt` and config set & not set
 - prep-node
       - with `--no-prompt` and config set & not set
       - without `--no-prompt` and config set & not set
       - with and without `--mfa`
 - attach-node
       - with `--no-prompt` and config set & not set
       - without `--no-prompt` and config set & not set
- all cloud provider check commands
       - with and without ``--no-prompt`
- bundle
       - with `--no-prompt` and config set & not set
       - without `--no-prompt` and config set & not set
- config set
       - with and without `--no-prompt`


cmd/config.go
```diff
@@ -3,284 +3,105 @@
...
-	Long:  `Create a new config that can be used to query Platform9 controller`,
-	Run:   configCmdCreateRun,
-	Args: func(configCmdSet *cobra.Command, args []string) error {
-		if configCmdSet.Flags().Changed("account_url") || configCmdSet.Flags().Changed("username") || configCmdSet.Flags().Changed("password") || configCmdSet.Flags().Changed("region") || configCmdSet.Flags().Changed("tenant") || configCmdSet.Flags().Changed("proxyUrl") {
-			SetConfigByParameters = true
-		}
-		return nil
-	},
-}
+const MaxLoopNoConfig = 3
```
nikhilbandi, line 67: Can this be put in constants file?

cmd/config.go
```diff
@@ -3,284 +3,105 @@
...
-)
-
-func init() {
-	configCmdSet.Flags().StringVarP(&account_url, "account_url", "u", "", "sets account_url")
-	configCmdSet.Flags().StringVarP(&username, "username", "e", "", "sets username")
-	configCmdSet.Flags().StringVarP(&Password, "password", "p", "", "sets password (use 'single quotes' to pass password)")
-	configCmdSet.Flags().StringVarP(&proxyURL, "proxy_url", "l", "", "sets proxy URL, can be specified as [<protocol>][<username>:<password>@]<host>:<port>")
-	configCmdSet.Flags().StringVarP(&region, "region", "r", "", "sets region")
-	configCmdSet.Flags().StringVarP(&tenant, "tenant", "t", "", "sets tenant")
+	configCmdSet.Flags().StringVarP(&cfg.Fqdn, "account_url", "u", "", "sets account_url")
```
nikhilbandi, line 74: Can't we have a common init() function somewhere instead of repeating this chunk of code in every file?

cmd/root.go
```diff
@@ -62,6 +63,7 @@ func init() {
 	cobra.OnInitialize(initConfig)
 
 	rootCmd.PersistentFlags().BoolVar(&verbosity, "verbose", false, "print verbose logs")
+	rootCmd.PersistentFlags().BoolVar(&detach, "dt", false, "disable all user prompts")
```
nikhilbandi, line 66: Is "dt" a commonly used abbreviation? Can we think of something more intuitive?

pkg/config/config.go
```diff
@@ -0,0 +1,426 @@
...
+
+}
+
+func GetConfigRecursive(loc string, cfg *objects.Config, nc objects.NodeConfig) error {
+	MaxLoopNoConfig := 3
+	InvalidExistingConfig := false
+	count := 0
+	var err error
+
+	if loc == "amazon.json" {
```
nikhilbandi, line 124: Shouldn't the json filenames be part of the constants file?

pkg/pmk/node.go
```diff
@@ -152,28 +154,28 @@ func PrepNode(ctx Config, allClients Client) error {
...
-	endpointURL, err := keystone.GetEndpointForRegion(ctx.Fqdn, auth, ctx.Region, regionInfoServiceID)
-	if err != nil {
-		return "", fmt.Errorf("Failed to fetch installer URL, Error: %s", err)
-	}
-	zap.S().Debug("endpointURL fetched : ", endpointURL)
-	return endpointURL, nil
-}
-
-func installHostAgent(ctx Config, auth keystone.KeystoneAuth, hostOS string, exec cmdexec.Executor) error {
+// func FetchRegionFQDN(ctx config.Config, auth keystone.KeystoneAuth) (string, error) {
```
nikhilbandi, line 157: This can be removed instead of just being commented.

cmd/config.go
sahil-lakhwani: This wasn't needed. Thanks for pointing.

cmd/config.go
sahil-lakhwani: I think it is better this way so that flags for respective commands remain local

pkg/config/config.go
sahil-lakhwani: This is something from existing code. We can remove this in a separate PR
I'm thinking to do this in a better way so - these file names are just dummy, no file with such names is even created

pkg/pmk/node.go
sahil-lakhwani: removed this and some more unused code, thanks for noticing+

cmd/attachNode.go
```diff
@@ -19,69 +23,64 @@ var (
...
-			zap.S().Debug("Error connecting to host %s", err.Error())
-			zap.S().Fatalf(" Invalid (Username/Password/IP)")
-		}
+	cfg := &objects.Config{WaitPeriod: time.Duration(60), AllowInsecure: false, MfaToken: attachconfig.MFA}
+	var err error
+	if detachedMode {
+		err = config.LoadConfig(util.Pf9DBLoc, cfg, nc)
+	} else {
+		err = config.LoadConfigInteractive(util.Pf9DBLoc, cfg, nc)
+	}
```
anupbarve, line 70: Can't this decision be made inside config module by passing detachedMode flag to a generic function ?
This will reduce the load on all the callers (existing and future ones) to invoke the same conditional logic.

cmd/checkNode.go
```diff
@@ -105,10 +103,17 @@ func checkNodeRun(cmd *cobra.Command, args []string) {
 		zap.S().Fatalf("Unable to obtain keystone credentials: %s", err.Error())
 	}
 
-	result, err := pmk.CheckNode(ctx, c, auth)
+	result, err := pmk.CheckNode(*cfg, c, auth)
+	if isRemote {
+		if err := SudoPasswordCheck(executor, detachedMode, nc.SudoPassword); err != nil {
+			zap.S().Fatal(err.Error())
```
anupbarve, line 109: Need to add the context of this fatal failure as a message.
Something like "Failed to check sudo password"

cmd/cloudProviders.go
```diff
@@ -94,62 +72,81 @@ func init() {
...
+	if cmd.Flags().Changed("dt") {
+		flagsNotSet := checkFlags(cmd)
+		if len(flagsNotSet) > 0 {
+			fmt.Printf(color.Red("x ")+"Missing required flags: %v\n", strings.Join(flagsNotSet, ", "))
 			os.Exit(1)
 		}
-		os.Exit(0)
+		err = config.LoadConfig("google.json", cfg, objects.NodeConfig{})
+	} else {
+		err = config.LoadConfigInteractive("google.json", cfg, objects.NodeConfig{})
```
anupbarve, line 85: Seems like the only difference in all the `*ProviderRun` functions is the json file. 
Could we factor out all the common code in those functions ?

cmd/prepNode_test.go
```diff
@@ -1,25 +0,0 @@
...
-
-	t.Run("LocalExecutorTest", func(t *testing.T) {
-		executor, err := getExecutor(proxy_url)
-		if err != nil {
-			t.Errorf("Error occured : %s", err)
-		}
-		assert.Equal(t, TestErr, err)
-		assert.Equal(t, exec, executor)
-	})
-}
```
anupbarve, line 25: Why deleted ? How can we replace this test if not valid anymore?

anupbarve (changes requested):

pkg/objects/objects.go
```diff
@@ -0,0 +1,36 @@
+package objects
```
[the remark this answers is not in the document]
sahil-lakhwani: Putting those structs in `config` package causes a import cycle.

cmd/attachNode.go
sahil-lakhwani: I think that is intentional to allow callers to do things before and after loading config
For example,
https://github.com/platform9/pf9ctl/blob/76a4d881f0454bd0dd256479d94ab85fd383e9df/cmd/cloudProviders.go#L77-L83

cmd/cloudProviders.go
sahil-lakhwani: Yes we can.  A lot of that is existing code. As mentioned above in a reply to Nikhil, I would like to take it in a different PR

cmd/prepNode_test.go
sahil-lakhwani: The functioned being tested has been moved. I'll add a test case for there.

pkg/keystone/keystone.go
```diff
@@ -129,3 +129,21 @@ func (k KeystoneImpl) GetAuth(
 		Email:     user["name"].(string),
 	}, nil
 }
+
+func FetchRegionFQDN(fqdn string, region string, auth KeystoneAuth) (string, error) {
```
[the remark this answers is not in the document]
sahil-lakhwani: Okay,I'll update this

cmd/checkNode.go
sahil-lakhwani: Added this

anupbarve (approved):
lgtm.
This is a relatively large change. Could you please document all the testing done in the PR description ?

PR #221, opened 2021-11-09
mfa fix

devidask27:
Testing done on ubuntu and centos
**check-node on ubuntu**
ub check-node
**prep-node on ubuntu**
Screenshot 2021-11-09 at 5 03 49 PM
**check-node on centos**
Screenshot 2021-11-09 at 2 28 51 PM
**prep-node on centos**
Screenshot 2021-11-09 at 2 27 29 PM

### DIFF
--- a/pkg/pmk/node.go
+++ b/pkg/pmk/node.go
@@ -55,23 +55,13 @@ func sendSegmentEvent(allClients Client, eventStr string, auth keystone.Keystone
 }
 
 // PrepNode sets up prerequisites for k8s stack
-func PrepNode(ctx Config, allClients Client) error {
+func PrepNode(ctx Config, allClients Client, auth keystone.KeystoneAuth) error {
 	// Building our new spinner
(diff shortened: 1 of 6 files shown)


pkg/pmk/node.go
```diff
@@ -55,23 +55,13 @@ func sendSegmentEvent(allClients Client, eventStr string, auth keystone.Keystone
 }
 
 // PrepNode sets up prerequisites for k8s stack
-func PrepNode(ctx Config, allClients Client) error {
+func PrepNode(ctx Config, allClients Client, auth keystone.KeystoneAuth) error {
```
nikhilbandi, line 58: Instead of passing the token to each required function, can we make it a global one so that it's accessible to all functions once package is included?

anupbarve (approved):
lgtm.

pkg/pmk/node.go
anupbarve: As discussed with nikhil, we will take this up as a separate task.

PR #224, opened 2021-11-10
Added segment event property to differentiate between OVA and non-OVA node

ShaunakJoshi1407:
- Added segment event property (infra) to differentiate between OVA and non-OVA node on the basis of presence of ovf.service file.

OVA-node Segment Event
Screenshot 2021-11-10 at 9 14 15 AM

Non OVA-node Segment Event
Screenshot 2021-11-10 at 9 14 23 AM
t


pkg/pmk/checkNode.go
```diff
@@ -35,6 +37,17 @@ var WarningOptionalChecks bool
...
+	//Checking for OVF Service
+	_, err := os.Stat(util.OVFLoc)
+	if err != nil {
+		zap.S().Debugf("OVF Service not present")
+		util.OvfServicePresent = false
+	} else {
+		zap.S().Debugf("Node onboarded through OVA")
+		util.OvfServicePresent = true
+	}
+
```
anupbarve, line 50: I don't think this is the right place to put the check.
Not every CLI command would use this path, so we will get selective events.

anupbarve (changes requested):

pkg/pmk/checkNode.go
ShaunakJoshi1407: Should I add this check in segment.go itself?

anupbarve (dismissed):

ShaunakJoshi1407:
Creating a seperate PR for OVA segemnt event property as this contains automation changes too

PR #236, opened 2021-11-23
Fixed decomission not removing node from ui

JasminAnt:
Fixed an error where the node would get deleted too fast from the project for the CLI to gets its uuid to delete it from the ui.


ShaunakJoshi1407 (approved):
Checked decommission-node command with --mfa flag on Ubuntu and Centos. It is working correctly in both the cases

PR #243, opened 2021-11-26
Fix for empty extensions in database

JasminAnt:
Fixed an error where if a user had an empty extensions string in databse it fetching node uid would not work.
image

image


nikhilbandi (approved):
Looks good.

PR #250, opened 2021-12-01
Fixed cluster removing from nodes

JasminAnt:
The decommission command was not killing cluster services and thus check-node would read as if the cluster was still on the node.
Deco1
Deco2
Deco3

Delete1
Delete2

Warns the user if the detached node is the last or a master node of a cluster
DetachNew


nikhilbandi (approved):
Changes look good.

PR #251, opened 2021-12-02
Added removing /opt dir

JasminAnt:
Decommission command was leaving /opt dir behind

image

image
...
image

### DIFF
--- a/cmd/decomissionNode.go
+++ b/cmd/decomissionNode.go
@@ -116,6 +116,8 @@ func decommissionNodeRun(cmd *cobra.Command, args []string) {
 		RunCommandWait("sudo pkill -9 `pidof etcd`")
 		RunCommandWait("sudo pkill -9 `pidof kube_proxy`")
 
+		RunCommandWait("sudo rm -rf /opt")
+
 	} else {
 		//command = "sudo yum erase -y pf9-hostagent -y"
 		fmt.Println("Removing packages")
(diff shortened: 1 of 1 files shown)


cmd/decomissionNode.go
```diff
@@ -116,6 +116,8 @@ func decommissionNodeRun(cmd *cobra.Command, args []string) {
 		RunCommandWait("sudo pkill -9 `pidof etcd`")
 		RunCommandWait("sudo pkill -9 `pidof kube_proxy`")
 
+		RunCommandWait("sudo rm -rf /opt")
```
anupbarve, line 119: I don't think we should be doing this.
This is removal of generic directory that might be used by the customer for other add-ons.

nikhilbandi: @JasminAnt , can you please confirm the removal of subfolders under /opt with the PMK team once?

JasminAnt: image

JasminAnt: @nikhilbandi  who can I message from the team about this? I am not in the slack channel so I dont know who is on that team

cmd/decomissionNode.go
```diff
@@ -130,12 +125,21 @@ func decommissionNodeRun(cmd *cobra.Command, args []string) {
...
-		RunCommandWait("sudo pkill -9 `pidof kubelet`")
-		RunCommandWait("sudo pkill -9 `pidof etcd`")
-		RunCommandWait("sudo pkill -9 `pidof kube_proxy`")
-
 	}
 
+	RunCommandWait("sudo pkill kubelet")
+	RunCommandWait("sudo pkill etcd")
+	RunCommandWait("sudo pkill kube-proxy")
+	RunCommandWait("sudo pkill kube-apiserve")
```
nikhilbandi, line 133: Isn't it kube-apiserver?

nikhilbandi (approved):
Approving.

cmd/decomissionNode.go
JasminAnt: image
Am getting apiserve from netstat

PR #254, opened 2021-12-09
made flag styling consistent

devidask27:


README.md
```diff
@@ -443,7 +443,6 @@ Usage:
 Flags:
   -h, --help   help for deauthorize-node
       --mfa string   MFA token
-  -s, --skipCheck    skips node check
```
nikhilbandi, line 446: Don't we have to replace this with --skip-check?

devidask27: This flag does not exist currently in deauthorize-node command. so removed it

nikhilbandi (changes requested):
Added 1 minor comment.

README.md
nikhilbandi: OK.

nikhilbandi (approved):
Approving.

PR #257, opened 2021-12-20
Remove .DS_Store

erwinvaneyk:


anupbarve (commented):
Maybe this should be added in .gitignore ?

erwinvaneyk:
@anupbarve I think most of the time it is something to set in the global gitignore. It also still requires removing the already checked in one

anupbarve:
@erwinvaneyk I was suggesting to add it at https://github.com/platform9/pf9ctl/blob/master/.gitignore so that it doesn't get checked in again.
There is no denying that we need to remove already existing one that has been checked in.

anupbarve (approved):

PR #258, opened 2022-01-06
fixed prep-node successful segment event

devidask27:
All events were getting send to segment except two events authorising host and prep-node successful events.

Before

Screenshot 2022-01-06 at 1 47 55 PM

After

Screenshot 2022-01-06 at 1 30 16 PM


anupbarve (commented):
can you please post testing info in description.

anupbarve (approved):

PR #261, opened 2022-01-11
Create remote directory during UploadFile

xagent003:
While trying new change for creating Comms registry.json file, it works on upgrade. But failing on new DU with fresh hosts. The /etc/pf9/comms/ dir does not exist. So UploadFile is failing. Might need changes in pf9ctl to have the sftpClient do a MkdirAll on the filepath's dir, before it does the remote .Create:

### DIFF
--- a/pkg/ssh/sshclient.go
+++ b/pkg/ssh/sshclient.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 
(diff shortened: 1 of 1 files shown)


mithilarun (approved):
These changes look good. Could you add similar code to the `DownloadFile` function below too?

nikhilbandi:
@xagent003 , just wanted to know if this change should be part of the upcoming pf9ctl release. 
CC: @anupbarve, @devidask27

PR #263, opened 2022-01-18
failure while sending segment event should not print error msg on console

devidask27:
Testing
**Before**
Screenshot 2022-01-18 at 3 15 17 PM
**Now**
Screenshot 2022-01-18 at 5 23 15 PM
**Logs**
Screenshot 2022-01-18 at 5 11 36 PM
Screenshot 2022-01-18 at 5 12 13 PM

### DIFF
--- a/pkg/client/segment.go
+++ b/pkg/client/segment.go
@@ -15,7 +15,7 @@ import (
 )
 
 //Added segment key for the source PRD-PMKFT Metrics-Aggregator
-var SegmentWriteKey = "4jevYUNBF5sY3vZJWm5TrhfdsFzQIQ3y"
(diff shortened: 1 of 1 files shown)


pkg/client/segment.go
```diff
@@ -120,6 +125,13 @@ func (c SegmentImpl) Close() {
 	c.client.Close()
 }
 
+func (c *SegmentNoopLogger) Logf(format string, args ...interface{}) {
+	zap.S().Debug("Cloud not send segment event")
```
anupbarve, line 129: %s/Cloud/Could
Here and below as well

anupbarve (approved):

PR #266, opened 2022-01-20
Private/devidas/attach node spinner

devidask27:
Testing
**Attaching 3 workers**
Screenshot 2022-01-20 at 4 43 06 PM
**Attaching 3 workers which are already connected to cluster**
Screenshot 2022-01-20 at 4 44 23 PM
**Attaching 3 workers out of three one is already connected to cluster**
Screenshot 2022-01-20 at 4 46 23 PM
**Attaching 3 workers out of three 2 are already connected to cluster**
Screenshot 2022-01-20 at 4 47 22 PM
**Attaching single worker node**
Screenshot 2022-01-20 at 4 49 12 PM
**Attaching single master node**
Screenshot 2022-01-20 at 5 40 22 PM
**Attaching 3 node as master which are already connected to cluster**
Screenshot 2022-01-20 at 5 04 56 PM
**Attaching one master two workers**
Screenshot 2022-01-20 at 5 08 33 PM
**When cluster is not ready**
Screenshot 2022-01-20 at 5 09 25 PM

### DIFF
--- a/cmd/attachNode.go
+++ b/cmd/attachNode.go
@@ -149,6 +149,8 @@ func attachNodeRun(cmd *cobra.Command, args []string) {
 					}
 					zap.S().Infof("Worker node(s) %v attached to cluster", wokerids)
 				}
+			} else {
+				zap.S().Infof("All workers are already attached. No worker node available to attach to the cluster")
 			}
 
 		}
@@ -177,6 +179,8 @@ func attachNodeRun(cmd *cobra.Command, args []string) {
 					}
 					zap.S().Infof("Master node(s) %v attached to cluster", masterids)
 				}
+			} else {
+				zap.S().Infof("All masters are already attached. No master node available to attach to the cluster")
 			}
 
 		}
@@ -218,6 +222,7 @@ func cluster_Status(exec cmdexec.Executor, fqdn string, token string, projectID 
 	return status
 }
 
+//Check if the node being attached is already attached to any cluster
 func isConnectedToAnyCluster(exec cmdexec.Executor, fqdn string, token string, projectID string, hostId string) string {
 	zap.S().Debug("Checking if node is connected to any cluster")
 	tkn := fmt.Sprintf(`"X-Auth-Token: %v"`, token)
--- a/cmd/attachNode.go
+++ b/cmd/attachNode.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/briandowns/spinner"
 	"github.com/platform9/pf9ctl/pkg/client"
 	"github.com/platform9/pf9ctl/pkg/cmdexec"
 	"github.com/platform9/pf9ctl/pkg/color"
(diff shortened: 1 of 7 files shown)


cmd/attachNode.go
```diff
@@ -118,39 +118,67 @@ func attachNodeRun(cmd *cobra.Command, args []string) {
...
+					}
+					zap.S().Info("Encountered an error while attaching worker node to a Kubernetes cluster : ", err1)
+				} else {
+					if err := c.Segment.SendEvent("Attaching-node", auth, "Worker node attached", ""); err != nil {
+						zap.S().Errorf("Unable to send Segment event for attach node. Error: %s", err.Error())
+					}
+					zap.S().Infof("Worker node(s) %v attached to cluster", wokerids)
 				}
-				zap.S().Infof("Worker node(s) %v attached to cluster", worker_hostIds)
 			}
```
nikhilbandi, line 152: Shouldn't we have else condition where we print console message and log saying "no worker node available to attach to the cluster"?

cmd/attachNode.go
```diff
@@ -118,39 +118,67 @@ func attachNodeRun(cmd *cobra.Command, args []string) {
...
+					}
+					zap.S().Info("Encountered an error while attaching master node to a Kubernetes cluster : ", err1)
+				} else {
+					if err := c.Segment.SendEvent("Attaching-node", auth, "Master node attached", ""); err != nil {
+						zap.S().Errorf("Unable to send Segment event for attach node. Error: %s", err.Error())
+					}
+					zap.S().Infof("Master node(s) %v attached to cluster", masterids)
 				}
-				zap.S().Infof("Master node(s) %v attached to cluster", master_hostIds)
 			}
```
nikhilbandi, line 180: Same as above, just for Master nodes. Shouldn't the relevant message be printed in the else condition?

cmd/attachNode.go
```diff
@@ -189,3 +217,15 @@ func cluster_Status(exec cmdexec.Executor, fqdn string, token string, projectID
 	zap.S().Debug("Cluster status is : ", status)
 	return status
 }
+
```
nikhilbandi, line 220: Add comment "//Check if the node being attached is already attached to any cluster."

nikhilbandi (changes requested):
Please take a look at the comments mentioned.

cmd/attachNode.go
devidask27: No Nikhil, we are first printing if it is attached to any cluster or not. So there is no node means that are already connected.

devidask27: Done

cmd/attachNode.go
devidask27: Done

cmd/attachNode.go
devidask27: Done

nikhilbandi (approved):
Approving.

PR #273, opened 2022-01-31
Providing Tenant UUID in pf9ctl config set results in an "Invalid Credentials" error

Shubhamag12:
Works fine when ID or name is passed.
testing_output

Works fine a random ID is passed.
Screenshot 2022-02-01 at 6 13 35 PM

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/briandowns/spinner v1.13.0
 	github.com/fatih/color v1.10.0
 	github.com/google/go-cmp v0.5.6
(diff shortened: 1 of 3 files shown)


devidask27:
Before merging do the testing for negative cases, Also test with mfa token.

Shubhamag12:
Checked for the negative cases too, here is the snapshot of case when MFA token is passed with ID or name
Screenshot 2022-02-01 at 5 16 25 PM
Screenshot 2022-02-01 at 5 13 51 PM

PR #276, opened 2022-02-03
Remove option of existing packages on re-run

joey00072:
Option to remove existing packages to commands check-node,prep-node
and bootstrap
Moved pf9ctl downloaded packages removing commands to pmk/decomissionNode
Moved func hostId,getIp to pmk/network to avoid circular import


pkg/pmk/checkNode.go
```diff
@@ -120,6 +127,23 @@ func CheckNode(ctx objects.Config, allClients client.Client, auth keystone.Keyst
...
+	removeCurrentInstallation := ""
+	if !cleanInstallCheck {
+		fmt.Println(color.Yellow("\n\nPrevious installation found"))
+		fmt.Println(color.Yellow("Reinstall Required..."))
+		fmt.Print("Remove Current Installation Type ('yes'/'no'):")
+		fmt.Scanf("%s", &removeCurrentInstallation)
+
+		if strings.ToLower(removeCurrentInstallation) == "yes" {
+
+			fmt.Print("Decomissing Node ...\n\n")
```
anupbarve, line 139: nit:
%s/Decomissing/Decommissioning

pkg/pmk/decomissionNode.go
```diff
@@ -0,0 +1,106 @@
...
+func RunCommandWait(command string) {
+	output := exec.Command("/bin/sh", "-c", command)
+	output.Stdout = os.Stdout
+	output.Stdin = os.Stdin
+	err := output.Start()
+	output.Wait()
+	if err != nil {
+		fmt.Println(err.Error())
+	}
+}
```
anupbarve, line 25: Do you still need `RunCommandWait` in cmd/decomissionNode.go ?

anupbarve (changes requested):
Please post screenshots of the testing done.
@devidask27 you should also review this change since you own the CLI code.

joey00072:
When irevious  installation  exists 

Screenshot 2022-02-08 at 6 53 52 PM

Clean install

Screenshot 2022-02-08 at 6 55 40 PM

anupbarve (changes requested):
I realised that this change will break automation.
Any prompt on terminal should also have corresponding flag on CLI to make decisions.
Example: Prompting for removal of packages should also have corresponding flag to be passed like `removeExistingPkgs`. If the flag is passed, no need to prompt, instead just make the decision based on passed value.

joey00072:
### Added flag to ignore remove existing pkgs prompt

Screenshot 2022-02-21 at 2 50 02 PM

cmd/bootstrap.go
```diff
@@ -151,7 +152,7 @@ func bootstrapCmdRun(cmd *cobra.Command, args []string) {
 	if !util.SkipPrepNode {
 		zap.S().Debug("========== Running check-node as a part of bootstrap ==========")
 
-		result, err := pmk.CheckNode(*cfg, c, auth, bootConfig)
+		result, err := pmk.CheckNode(*cfg, c, auth, bootConfig, removeExistingPkgs)
```
anupbarve, line 155: see if you can fold the flag `removeExistingPkgs` in config structure.
Passing every new added flag via function params is a pain and should be avoided as much as possible.

anupbarve (changes requested):

anupbarve (approved):
lgtm
Please do an end-to-end testing and also get approval from @devidask27

devidask27:
@joey00072 Please post the all the latest testing screenshots for all the possible positive and negative test cases.

devidask27:
@joey00072 Also test the other functionalities for authorize-node, deauthorize-node, looks like this files also changed because of conflicts. so just test the functionality of that changes as well.

devidask27:
@joey00072 By looking at the code, it looks it is not handling the case where the node is partially onboarded and we enter no when it ask for do you want to remove existing installation. I think instead of asking user to enter yes or no we should directly remove the existing installation if user is re-running the prep-node. CC: @anupbarve

joey00072:
> @joey00072 By looking at the code, it looks it is not handling the case where the node is partially onboarded and we enter no when it ask for do you want to remove existing installation. I think instead of asking user to enter yes or no we should directly remove the existing installation if user is re-running the prep-node. CC: @anupbarve

Shouldn't we ask for user permission before deleting the packages from host machine? 
@devidask27 @anupbarve

joey00072:
### Ubuntu

#### prep-node no
ubuntu-prep-node-no

#### prep-node yes
ubuntu-prep-node-yes

#### prep-node with flag
ubuntu-prep-node-with-flag

#### check-node no
ubuntu-check-node-no


#### check-node yes
ubuntu-check-node-yes

#### check-node with flag
ubuntu-check-node-with-flag



### Centos

#### prep-node no
centos-prep-node-no


#### prep-node yes
centos-prep-node-yes_1
centos-prep-node-yes_2



#### prep-node with flag
centos-prep-node-with-flag-1
centos-prep-node-with-flag-2

#### check-node no
centos-check-node-no



#### check-node yes
centos-check-node-yes-1
centos-check-node-yes-2

#### check-node with flag
centos-check-node-with-flag-2

centos-check-node-with-flag-1

PR #291, opened 2022-03-28
added flag to pass log-location and also logged the console messages into log file

devidask27:
## ISSUE(S):

https://platform9.atlassian.net/browse/FT-262

## SUMMARY

Added new flag to pass location to save logs, and also logged the console messages into logfile

## ISSUE TYPE

## RELATED ISSUE(S):

https://platform9.atlassian.net/browse/FT-361

#### Manual

**Flag to pass log location**
Screenshot 2022-03-28 at 12 57 48 PM
**User defined log location to save logs**
Screenshot 2022-03-28 at 12 59 21 PM
**Default log location**
Screenshot 2022-03-28 at 1 02 07 PM

#### Reviewers

/cc @anupbarve @nikhilbandi

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -86,6 +86,14 @@ func initConfig() {
 	}*/
 
 	if rootCmd.Flags().Changed("log-dir") {
+		logDirPath = filepath.Join(logDirPath)
+		if _, err := os.Stat(logDirPath); os.IsNotExist(err) {
+			zap.S().Debugf("%s dir is not preset creating it", logDirPath)
+			err = os.MkdirAll(logDirPath, 0700)
+			if err != nil {
+				zap.S().Fatalf("error creating %s dir, please make sure dir is present")
+			}
+		}
 		util.Pf9Log = filepath.Join(logDirPath, "pf9ctl.log")
 		util.Pf9LogLoc = logDirPath
 	} else {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -89,7 +89,7 @@ func initConfig() {
 		util.Pf9Log = filepath.Join(logDirPath, "pf9ctl.log")
 		util.Pf9LogLoc = logDirPath
 	} else {
-		util.Pf9LogLoc = "pf9/log"
+		util.Pf9LogLoc = util.DefaultPf9LogLoc
 	}
 
 	// Read in environment variables that match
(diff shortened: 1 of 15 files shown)


devidask27:
Log file
[pf9ctl-20220328.log](https://github.com/platform9/pf9ctl/files/8361663/pf9ctl-20220328.log)

cmd/root.go
```diff
@@ -82,6 +85,13 @@ func initConfig() {
 		viper.SetConfigName(".pf9ctl")
 	}*/
 
+	if rootCmd.Flags().Changed("log-dir") {
+		util.Pf9Log = filepath.Join(logDirPath, "pf9ctl.log")
+		util.Pf9LogLoc = logDirPath
+	} else {
```
anupbarve, line 91: What happens when the logDirPath doesn't exist ?
The directory should be created with appropriate permissions.

cmd/root.go
```diff
@@ -82,6 +85,13 @@ func initConfig() {
...
+		util.Pf9LogLoc = "pf9/log"
```
anupbarve, line 92: Add to constants.

pkg/util/constants.go
```diff
@@ -95,7 +95,7 @@ var (
 	OVFLoc    = "/etc/systemd/system/ovf.service"
 	VarDir    = "/var/log/pf9"
 	EtcDir    = "/etc/pf9"
-	Pf9LogLoc = "pf9/log"
+	Pf9LogLoc string
```
anupbarve, line 98: why ?

anupbarve (changes requested):

cmd/root.go
devidask27: if the logDirPath doesn't exist then it throughs error with message no such file or directory
Screenshot 2022-04-04 at 10 37 32 AM

pkg/util/constants.go
devidask27: Previously default log locations was "pf9/log" but now its depends on user so assigning value based on flag.

cmd/root.go
devidask27: Done

cmd/root.go
anupbarve: We need to create the directory via code as mentioned above instead of erroring out.

anupbarve (approved):

PR #296, opened 2022-04-18
Private/shubhamc/cli dpkg lock check failed ft 336

joey00072:
## ISSUE(S):

CLI dpkg lock check failed. Because  of unattended-upgrades
[FT-366](https://platform9.atlassian.net/browse/FT-366)

## SUMMARY

If unattended-upgrades start after dpkg lock check in check-node.
dpkg would go under lock and we would unable to install hostagent.

Added additional dpkg lock check after stoping unattended-upgrades.
Also add check if unattended-upgrades is disabled by default.

## ISSUE TYPE

changed debian.checkIfdpkgISLock to debian.CheckIfdpkgISLock to use outside debian package

#### If unattended-upgrades starts running before check-node
Screenshot 2022-04-11 at 2 50 28 PM

#### If unattended-upgrades starts running after check-node
Screenshot 2022-04-11 at 3 59 42 PM

#### If unattended-upgrades active but not running yet
Screenshot 2022-04-14 at 12 18 36 PM

#### Reviewers
@devidask27 @AnirudhPokala @anupbarve @yogeshj83

### DIFF
--- a/pkg/platform/debian/debian.go
+++ b/pkg/platform/debian/debian.go
@@ -331,6 +331,26 @@ func (d *Debian) disableSwap() (bool, error) {
 	}
 }
 
+func (d *Debian) processAcquiredDpkgLock() (string, error) {
+	var f = []string{"lock", "lock-frontend"}
+	for _, file := range f {
+
+		output, err := d.exec.RunWithStdout("bash", "-c", fmt.Sprintf("lsof /var/lib/dpkg/%s | grep lib/dpkg/%s ", file, file))
+		if err != nil {
+			return "", errors.New("Unable to find pid with dpkg lock")
+		}
+		PID := strings.Fields(output)[1]
+		output, err = d.exec.RunWithStdout("bash", "-c", fmt.Sprintf("ps -p %s -o command=", PID))
+		if err != nil {
+			return "", errors.New("Unable to find process with dpkg lock")
+		}
+		output = strings.Replace(output, "\n", "", -1)
+		return fmt.Sprintf("Process: '%s', Pid: %s ", output, PID), nil
+	}
+	return "", errors.New("Unable to find process with dpkg lock")
+
+}
+
 func (d *Debian) CheckIfdpkgISLock() (bool, error) {
 	var f = []string{"lock", "lock-frontend"}
 	for _, file := range f {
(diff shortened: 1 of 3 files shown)


joey00072:
#### When dpkg locked by unattended-upgrade
Screenshot 2022-04-25 at 4 37 04 PM

#### When dpkg lock by apt
Screenshot 2022-04-25 at 4 37 42 PM

pkg/pmk/node.go
```diff
@@ -75,9 +75,26 @@ func PrepNode(ctx objects.Config, allClients client.Client, auth keystone.Keysto
 		return fmt.Errorf(errStr)
 	}
 
+	platform := debian.NewDebian(allClients.Executor)
```
devidask27, line 78: I think this will fail in case of centos. this should be called if and only if host os is debian.

devidask27 (changes requested):

pkg/pmk/node.go
joey00072: Fixed 👍

devidask27 (dismissed):
Looks good.

pkg/platform/debian/debian.go
```diff
@@ -331,16 +331,41 @@ func (d *Debian) disableSwap() (bool, error) {
...
 	var f = []string{"lock", "lock-frontend"}
 	for _, file := range f {
 		_, err := d.exec.RunWithStdout("bash", "-c", fmt.Sprintf("lsof /var/lib/dpkg/%s", file))
 		if err != nil {
 			return true, nil
 		} else {
 			zap.S().Debugf("Unable to acquire the dpkg lock on %s", file)
-			return false, fmt.Errorf("Unable to acquire the dpkg")
+			output, _ := d.processAcquiredDpkgLock()
+			return false, fmt.Errorf(fmt.Sprintf("Unable to acquire the dpkg - %s", output))
```
anupbarve, line 368: Isn't there a race condition where lock was acquired but got released and you try to print the process ID, but fail. This will fail the entire operation ?

joey00072: Screenshot 2022-04-28 at 2 00 50 PM

No In that case it will just return empty string and output will be like this

pkg/pmk/node.go
```diff
@@ -150,21 +168,61 @@ func PrepNode(ctx objects.Config, allClients client.Client, auth keystone.Keysto
 	return nil
 }
 
+func StatusUnattendedUpdates(allClients client.Client) bool {
+	zap.S().Debug("Checking Status of unattended-upgrades")
+	output, err := allClients.Executor.RunWithStdout("bash", "-c", "systemctl status unattended-upgrades | grep -i Active")
```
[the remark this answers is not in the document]
joey00072: Done 👍

anupbarve (approved):

PR #298, opened 2022-04-21
retry if status code is 409

devidask27:
## ISSUE(S):

https://platform9.atlassian.net/browse/FT-369
## SUMMARY

Ideally if API fails in first attempt then it should retry while node authorisation but it was not retrying for API status code 409 

## ISSUE TYPE

prep-node

#### Manual

Screenshot 2022-04-21 at 11 53 00 AM

### DIFF
--- a/pkg/util/helper.go
+++ b/pkg/util/helper.go
@@ -5,11 +5,12 @@ import (
 	"context"
 	"crypto/x509"
 	"fmt"
-	"go.uber.org/zap"
 	"net/http"
 	"net/url"
 	"os"
 	"regexp"
+
+	"go.uber.org/zap"
 )
 
 var (
@@ -58,7 +59,7 @@ func RetryPolicyOn404(ctx context.Context, resp *http.Response, err error) (bool
 	// 429 Too Many Requests is recoverable. Sometimes the server puts
 	// a Retry-After response header to indicate when the server is
 	// available to start processing request from client.
-	if resp.StatusCode == http.StatusTooManyRequests {
+	if resp.StatusCode == http.StatusTooManyRequests || resp.StatusCode == 409 {
 		return true, nil
 	}
 
--- a/pkg/util/helper.go
+++ b/pkg/util/helper.go
@@ -59,6 +59,8 @@ func RetryPolicyOn404(ctx context.Context, resp *http.Response, err error) (bool
 	// 429 Too Many Requests is recoverable. Sometimes the server puts
 	// a Retry-After response header to indicate when the server is
 	// available to start processing request from client.
+
+	//409 Resource conflicts. Most of the time this error(authorizing host) is transient so putting this into retry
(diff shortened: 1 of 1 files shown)


pkg/util/helper.go
```diff
@@ -5,11 +5,12 @@ import (
 	"context"
 	"crypto/x509"
 	"fmt"
-	"go.uber.org/zap"
 	"net/http"
 	"net/url"
 	"os"
 	"regexp"
+
```
nikhilbandi, line 12: Why this extra line?

pkg/util/helper.go
```diff
@@ -58,15 +59,15 @@ func RetryPolicyOn404(ctx context.Context, resp *http.Response, err error) (bool
 	// 429 Too Many Requests is recoverable. Sometimes the server puts
 	// a Retry-After response header to indicate when the server is
 	// available to start processing request from client.
-	if resp.StatusCode == http.StatusTooManyRequests {
+	if resp.StatusCode == http.StatusTooManyRequests || resp.StatusCode == http.StatusConflict {
```
nikhilbandi, line 62: Can you please add a comment for the 409 / StatusConflict error regarding why should it be retried?

pkg/util/helper.go
```diff
@@ -58,15 +59,15 @@ func RetryPolicyOn404(ctx context.Context, resp *http.Response, err error) (bool
...
 		return true, nil
 	}
 
 	// Check the response code. We retry on 500-range responses to allow
 	// the server time to recover, as 500's are typically not permanent
 	// errors and may relate to outages on the server side. This will catch
 	// invalid response codes as well, like 0 and 999.
-	if resp.StatusCode == 0 || resp.StatusCode == 400 || resp.StatusCode == 404 || (resp.StatusCode >= 500 && resp.StatusCode != 501) {
+	if resp.StatusCode == 0 || resp.StatusCode == http.StatusBadRequest || resp.StatusCode == http.StatusNotFound || (resp.StatusCode >= http.StatusInternalServerError && resp.StatusCode != http.StatusNotImplemented) {
```
nikhilbandi, line 70: Should we map the status messages to the error codes in comment, just for others to understand what error codes we are handling?
e.g. StatusBadRequest = 400

nikhilbandi (changes requested):
Provided minor suggestions. Please check.

pkg/util/helper.go
devidask27: Please ignore this. its because of editor.

pkg/util/helper.go
devidask27: I think no need, they are standard codes.

pkg/util/helper.go
devidask27: Done

nikhilbandi (approved):
Approving.

PR #308, opened 2022-05-27
unattended-upgrades disable issue

devidask27:
## ISSUE(S):

https://platform9.atlassian.net/browse/FT-388
## SUMMARY

Before running the installer script we are disabling the unattended-upgrades so that there should not be any problem while installing hostagent, but one of the user git dpkg lock error while installing hostagent.

## ISSUE TYPE

- [x] Breaking change (fix or feature that may cause existing functionality to not work as expected)

prep-node

#### Manual

**if user already have disabled unattended-upgrades**
Screenshot 2022-05-27 at 10 47 31 AM
**if unattended-upgrades are enabled**
Screenshot 2022-05-27 at 11 00 07 AM
Screenshot 2022-05-27 at 11 00 57 AM

#### Reviewers

/cc @anupbarve @anmolsachan

### DIFF
--- a/pkg/pmk/node.go
+++ b/pkg/pmk/node.go
@@ -79,22 +79,16 @@ func PrepNode(ctx objects.Config, allClients client.Client, auth keystone.Keysto
 
 		platform := debian.NewDebian(allClients.Executor)
 		result, _ := platform.CheckIfdpkgISLock()
-
-		if StatusUnattendedUpdates(allClients) {
-			// stop unattended-upgrades
-			// this do not stop them if they are already running
-			StopUnattendedUpdates(allClients)
-			defer StartUnattendedUpdates(allClients)
-
-			if IsEnabledUnattendedUpdates(allClients) {
-				DisableUnattendedUpdates(allClients)
-				defer EnableUnattendedUpdates(allClients)
-			}
-		}
-
 		if !result {
 			zap.S().Error("Dpkg lock is acquired by another process while prep-node was running")
 			return fmt.Errorf("Dpkg is under lock")
+		}
+
+		if ifUnattendedUpgradesAreEnabked(allClients) {
+			DisableUnattendedUpgrades(allClients)
+			defer EnableUnattendedUpgrades(allClients)
+		} else {
+			zap.S().Debug("unattended-upgrades are diabled by user")
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/pmk/node.go
```diff
@@ -168,64 +162,42 @@ func PrepNode(ctx objects.Config, allClients client.Client, auth keystone.Keysto
...
-	_, err := allClients.Executor.RunWithStdout("bash", "-c", "systemctl start unattended-upgrades")
-	if err != nil {
-		zap.S().Debugf("Failed to start unattended-upgrades : %s", err)
-	} else {
-		zap.S().Debug("Started unattended-upgrades")
-	}
-}
-
-func IsEnabledUnattendedUpdates(allClients client.Client) bool {
+func ifUnattendedUpgradesAreEnabked(allClients client.Client) bool {
```
ParagDoke, line 165: Same here

pkg/pmk/node.go
```diff
@@ -79,23 +79,17 @@ func PrepNode(ctx objects.Config, allClients client.Client, auth keystone.Keysto
...
-				defer EnableUnattendedUpdates(allClients)
-			}
-		}
-
 		if !result {
 			zap.S().Error("Dpkg lock is acquired by another process while prep-node was running")
 			return fmt.Errorf("Dpkg is under lock")
 		}
+
+		if ifUnattendedUpgradesAreEnabked(allClients) {
```
ParagDoke, line 87: ifUnattendedUpgradesAreEnabked -> ifUnattendedUpgradesAreEnabled
Typo

ParagDoke (changes requested):

pshanbhag:
@devidask27 - I am not for such a change. It is too intrusive and I don't think we are restoring the file back to the original settings. 
Is there value in this check? I would like to discuss with anyone who is pushing for this check. One option is to detect unattended upgrade is running and share the message with the user.

devidask27:
> @devidask27 - I am not for such a change. It is too intrusive and I don't think we are restoring the file back to the original settings. Is there value in this check? I would like to discuss with anyone who is pushing for this check. One option is to detect unattended upgrade is running and share the message with the user.

The issue is hostagent installation failed because of dpkg was acquired by unattended upgrades. we have check for dpkg lock check. it checks if dpkg lock is acquired by any other process. if it is acquired then it error out. but here the case is hostagent installation is failing if unattended upgrade or any other process acquires dpkg lock after our dpkg check or while hostagent is being installed.

PR #311, opened 2022-06-14
Create trivy.yml

cruizen:
Enable license scan with Trivy

## ISSUE TYPE

### DIFF
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -1,0 +1,47 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+name: build
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "master" ]
+  schedule:
+    - cron: '31 6 * * 3'
+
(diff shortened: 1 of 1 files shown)


cruizen:
TODO: This will need a Dockerfile.

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,47 @@
...
+permissions:
+  contents: read
+
+jobs:
+  build:
+    permissions:
+      contents: read # for actions/checkout to fetch code
+      security-events: write # for github/codeql-action/upload-sarif to upload SARIF results
+    name: Build
+    runs-on: "ubuntu-18.04"
```
NeelavaChatterjee, line 26: is there a requirement for this to be u18 can we use a higher one like u24?

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,47 @@
...
+name: build
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "master" ]
+  schedule:
+    - cron: '31 6 * * 3'
```
NeelavaChatterjee, line 15: This seems to be 6:31 every Wed, so does it need to be every week? We are already doing it one during PRs so maybe once a month would suffice?

PR #315, opened 2022-06-30
Add option to skip node auth in prep-node

pacharya-pf9:
## SUMMARY

PMK CAPI uses `pf9ctl prep-node` command in boot script of all instances. However pf9ctl always pushes the latest version to a node. This conflicts with the version being pushed by controller-manager. Although eventually the version pushed by controller-manager is one installed on the host, this process is delayed because pf9-kube package is downloaded, installed and removed at least couple of times.

This commits adds a new hidden option to NOT install pf9-kube role as part of the prep-node command. This hidden option will be used in the PMK CAPI project to remove the race condition since only one component i.e. controller-manager will push/remove the pf9-kube role from the host in case it is CAPI-managed.

## ISSUE TYPE

## RELATED ISSUE(S):

https://github.com/platform9/pf9-cluster-api/pull/137
https://github.com/platform9/pf9-qbert/pull/1348

### DIFF
--- a/cmd/prepNode.go
+++ b/cmd/prepNode.go
@@ -63,7 +63,7 @@ func init() {
 	prepNodeCmd.Flags().StringVar(&nodeConfig.MFA, "mfa", "", "MFA token")
 	prepNodeCmd.Flags().StringVarP(&nodeConfig.SudoPassword, "sudo-pass", "e", "", "sudo password for user on remote host")
 	prepNodeCmd.Flags().BoolVarP(&nodeConfig.RemoveExistingPkgs, "remove-existing-pkgs", "r", false, "Will remove previous installation if found (default false)")
-	prepNodeCmd.Flags().BoolVarP(&skipKube, "skip-kube", "", false, "Skip installing pf9-kube/nodelet on this host")
+	prepNodeCmd.Flags().BoolVarP(&util.SkipKube, "skip-kube", "", false, "Skip installing pf9-kube/nodelet on this host")
 	prepNodeCmd.Flags().MarkHidden("skip-kube")
 
 	rootCmd.AddCommand(prepNodeCmd)
@@ -170,7 +170,7 @@ func prepNodeRun(cmd *cobra.Command, args []string) {
 		}
 	}
 
-	if err := pmk.PrepNode(*cfg, c, auth, skipKube); err != nil {
+	if err := pmk.PrepNode(*cfg, c, auth); err != nil {
 
 		// Uploads pf9cli log bundle if prepnode failed to get prepared
 		errbundle := supportBundle.SupportBundleUpload(*cfg, c, isRemote)
--- a/cmd/prepNode.go
+++ b/cmd/prepNode.go
@@ -47,6 +47,7 @@ var (
 	ips            []string
 	skipChecks     bool
 	disableSwapOff bool
+	skipKube       bool = false
 )
 
 var nodeConfig objects.NodeConfig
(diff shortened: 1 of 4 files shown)


cmd/prepNode.go
```diff
@@ -47,6 +47,7 @@ var (
 	ips            []string
 	skipChecks     bool
 	disableSwapOff bool
+	skipKube       bool = false
```
devidask27, line 50: We can either make this exportable or we can put into constants.go (in util pkg). so that we don't need to change (pass one more value to prepNode function) definition of prepNode function.

devidask27 (changes requested):

cmd/prepNode.go
pacharya-pf9: Done.

erwinvaneyk, line 50: The `false` is not really needed

pkg/pmk/node.go
```diff
@@ -145,13 +145,18 @@ func PrepNode(ctx objects.Config, allClients client.Client, auth keystone.Keysto
 	s.Stop()
 	fmt.Println(color.Green("✓ ") + "Initialised host successfully")
 	zap.S().Debug("Initialised host successfully")
+	if util.SkipKube {
+		zap.S().Debug("Skip authorizing host as --skip-kube flag is true")
```
erwinvaneyk, line 149: Wouldn't it be more fitting to call this `--skip-auth`?

erwinvaneyk (approved):
Minor comment, lgtm in any case

pkg/pmk/node.go
pacharya-pf9: I intentionally kept it as --skip-kube as I don't want kube role to be installed. However if in future we add support for pushing aka authorizing other roles e.g. pf9-support then that would be out of scope of this flag.

cmd/prepNode.go
pacharya-pf9: Yeah... I was trying to be explicit... Dropped that `= false`

cmd/prepNode.go
```diff
@@ -58,10 +59,12 @@ func init() {
 	prepNodeCmd.Flags().StringSliceVarP(&nodeConfig.IPs, "ip", "i", []string{}, "IP address of host to be prepared")
 	prepNodeCmd.Flags().BoolVarP(&skipChecks, "skip-checks", "c", false, "Will skip optional checks if true")
 	prepNodeCmd.Flags().BoolVarP(&disableSwapOff, "disable-swapoff", "d", false, "Will skip swapoff")
-	prepNodeCmd.Flags().StringVar(&nodeConfig.MFA, "mfa", "", "MFA token")
 	prepNodeCmd.Flags().MarkHidden("disable-swapoff")
+	prepNodeCmd.Flags().StringVar(&nodeConfig.MFA, "mfa", "", "MFA token")
 	prepNodeCmd.Flags().StringVarP(&nodeConfig.SudoPassword, "sudo-pass", "e", "", "sudo password for user on remote host")
 	prepNodeCmd.Flags().BoolVarP(&nodeConfig.RemoveExistingPkgs, "remove-existing-pkgs", "r", false, "Will remove previous installation if found (default false)")
+	prepNodeCmd.Flags().BoolVarP(&util.SkipKube, "skip-kube", "", false, "Skip installing pf9-kube/nodelet on this host")
```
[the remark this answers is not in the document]
pacharya-pf9: done

devidask27 (approved):
lgtm

PR #325, opened 2022-08-30
skipping prep-node if node is already onboarded

devidask27:
## SUMMARY

Introduced new flag for prep-node command to skip prep-node if node is already connected to DU
## ISSUE TYPE

prep-node

#### Manual

Screenshot 2022-08-30 at 5 10 22 PM

#### Reviewers

/cc @pshanbhag  @anmolsachan

### DIFF
--- a/pkg/pmk/checkNode.go
+++ b/pkg/pmk/checkNode.go
@@ -4,6 +4,7 @@ package pmk
 
 import (
 	"fmt"
+	"os"
 	"strings"
 	"time"
 
@@ -15,6 +16,7 @@ import (
 	"github.com/platform9/pf9ctl/pkg/platform"
 	"github.com/platform9/pf9ctl/pkg/platform/centos"
 	"github.com/platform9/pf9ctl/pkg/platform/debian"
+	"github.com/platform9/pf9ctl/pkg/util"
 	"go.uber.org/zap"
 )
 
@@ -32,7 +34,9 @@ const (
 	CleanInstallFail CheckNodeResult = "cleanInstallFail"
 )
 
-/* This flag is set true, to have warning "!" message,
+/*
+	This flag is set true, to have warning "!" message,
+
 when user passes --skip-checks and optional checks fails.
 */
 var WarningOptionalChecks bool
@@ -49,13 +53,13 @@ func CheckNode(ctx objects.Config, allClients client.Client, auth keystone.Keyst
 	if !isSudo {
 		return RequiredFail, fmt.Errorf("User executing this CLI is not allowed to switch to privileged (sudo) mode")
 	}
-	os, err := ValidatePlatform(allClients.Executor)
+	hostOS, err := ValidatePlatform(allClients.Executor)
 	if err != nil {
 		return RequiredFail, err
 	}
 
 	var platform platform.Platform
-	switch os {
+	switch hostOS {
 	case "debian":
 		platform = debian.NewDebian(allClients.Executor)
 	case "redhat":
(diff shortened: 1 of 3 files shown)


pkg/pmk/checkNode.go
```diff
@@ -75,6 +79,54 @@ func CheckNode(ctx objects.Config, allClients client.Client, auth keystone.Keyst
...
+					//If node is already to connected then resmgr will return hostID
+					//If hostID is empty then host could be connected to other DU
+					var connected bool
+					if len(id) != 0 {
+						connected = allClients.Resmgr.HostSatus(auth.Token, id[0])
+					} else {
+						zap.S().Infof("Hostagent is installed on this host, but this host is not part of the DU %s specified in the config", ctx.Fqdn)
+						os.Exit(0)
+					}
+					if connected {
```
pshanbhag, line 106: What happens if the node is connected with hostagent but there is no role authorized?

pkg/pmk/checkNode.go
```diff
@@ -75,6 +79,54 @@ func CheckNode(ctx objects.Config, allClients client.Client, auth keystone.Keyst
...
+						zap.S().Info(color.Green("✓ ") + "Node is already connected\n")
+						os.Exit(0)
+					} else {
+						//case where hostagent is installed but host is in disconnected sate
+						zap.S().Debug("Hostagent is installed but host is not connected to DU, Trying to authorize")
+						if id[0] != "" {
+							err := allClients.Resmgr.AuthorizeHost(id[0], auth.Token)
+							if err != nil {
+								zap.S().Debug("Failed to authorize node, installing hostagent again")
+								nc.RemoveExistingPkgs = true
```
pshanbhag, line 117: Can you explain why we are trying to do this auth if the node is detected as not connected?

pshanbhag (commented):
@anmolsachan @devidask27 - what should be the behavior if the node is already authorized to the DU?

pkg/pmk/checkNode.go
devidask27: > What happens if the node is connected with hostagent but there is no role authorized?

In this case it will try to authorize node. And if authorization fails then it remove previous installation and install it again.

pkg/pmk/checkNode.go
devidask27: As @anmolsachan  stated. User is using CLI in script where he is giving let say 10 nodes and after 5 nodes CLI failed for some node. In this case user is running script again and this time CLI is prompting for yes/no to remove existing installation so user wants functionality to exclude nodes which are already connected.

Now coming to question why we are trying to do this auth if the node is detected as not connected?
In case if hostagent installed on host but it is not connected to DU so instead of reinstalling it again and then authorizing it to DU. Just trying to authorize it and if fails to authorize then removing existing installation and calling prep-node again.

pkg/pmk/checkNode.go
```diff
@@ -75,6 +79,54 @@ func CheckNode(ctx objects.Config, allClients client.Client, auth keystone.Keyst
...
+					ip = strings.TrimSpace(ip)
+					var add = []string{ip}
+					id := allClients.Resmgr.GetHostId(auth.Token, add)
+					//If node is already to connected then resmgr will return hostID
+					//If hostID is empty then host could be connected to other DU
+					var connected bool
+					if len(id) != 0 {
+						connected = allClients.Resmgr.HostSatus(auth.Token, id[0])
+					} else {
+						zap.S().Infof("Hostagent is installed on this host, but this host is not part of the DU %s specified in the config", ctx.Fqdn)
```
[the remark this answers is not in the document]
devidask27: Done

pkg/pmk/checkNode.go
pshanbhag: Won't the os.exit(0) on line 108 lead to no operations being performed?

pkg/pmk/checkNode.go
pshanbhag: But does authorizing bring it online? Isn't it a concern that you have an offline host that you are trying authorize?

I am coming from a point of view of simplicity... we need the CLI to do a fixed set of predictable operations. Keeping it simple goes a long way in maintenance. I am not sure these variations help unless you have seen cases where this would make a difference.

devidask27: yes authorizing host will bring it online in case hostagent is already installed and becuase of some reason host is not connected.

devidask27: I think here we can remove this authorizing host step if it is not connected. lets not make it complicated. if host is not connected will simply run prep-node again

anmolsachan:
@pshanbhag @devidask27 Is this good enough to release now?

devidask27:
> @pshanbhag @devidask27 Is this good enough to release now?

Yes. Waiting for approval.

pkg/pmk/checkNode.go
```diff
@@ -75,6 +79,44 @@ func CheckNode(ctx objects.Config, allClients client.Client, auth keystone.Keyst
...
+					}
+					ip = strings.Split(ip, " ")[0]
+					ip = strings.TrimSpace(ip)
+					var add = []string{ip}
+					id := allClients.Resmgr.GetHostId(auth.Token, add)
+					//If node is already connected then resmgr will return hostID
+					//If hostID is empty then host could be connected to other DU
+					var connected bool
+					if len(id) != 0 {
+						connected = allClients.Resmgr.HostSatus(auth.Token, id[0])
```
pshanbhag, line 101: At some point, you should fix this typo in `HostSatus`

pkg/pmk/checkNode.go
pshanbhag: Is my understanding correct on os.exit(0) implementation?

pshanbhag (approved):

pkg/pmk/checkNode.go
devidask27: > Is my understanding correct on os.exit(0) implementation?

yes os.exit(0) leads to no operation.

pkg/pmk/checkNode.go
devidask27: yes will take care.

PR #330, opened 2022-09-20
if failed to install required os pkgs return error

devidask27:
## SUMMARY

returning error on failing of required os packages installation

## ISSUE TYPE

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that may cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## IMPACTED FEATURES/COMPONENTS:

check-node

#### Manual

Screenshot 2022-09-20 at 3 09 01 PM

#### Reviewers

/cc @anmolsachan @pshanbhag

PR #331, opened 2022-09-23
Revert "if failed to install required os pkgs return error"

pshanbhag:
Reverts platform9/pf9ctl#330


pshanbhag:
An actual fix is being worked in in #332

PR #335, opened 2022-11-15
Add --kube-version hidden flag to prep node

pacharya-pf9:
This commit adds a --kube-version arg to prep node command. This will be used in CAPI boot scripts to simplifying and improve the onboarding process. Currently marking this as a hidden flag but if this turns out to be useful it can be made visible to the users as well.

## ISSUE(S):

[PMK-5344](https://platform9.atlassian.net/browse/PMK-5244): Support running prep-node with an explicit kubeversion in pf9ctl

## ISSUE TYPE

#### Manual

In-progress

### DIFF
--- a/pkg/resmgr/resmgr.go
+++ b/pkg/resmgr/resmgr.go
@@ -59,7 +59,7 @@ func (c *ResmgrImpl) AuthorizeHost(hostID string, token string, version string) 
 	url := fmt.Sprintf("%s/resmgr/v1/hosts/%s/roles/pf9-kube", c.fqdn, hostID)
 	if len(version) != 0 {
 		// Only resmgr v2 APIs support specifying a specific version of the role to install
-		url = fmt.Sprintf("%s/resmgr/v2/hosts/%s/roles/pf9-kube/?version=%s", c.fqdn, hostID, version)
+		url = fmt.Sprintf("%s/resmgr/v2/hosts/%s/roles/pf9-kube/versions/%s", c.fqdn, hostID, version)
 	}
 	req, err := rhttp.NewRequest("PUT", url, nil)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


erwinvaneyk (dismissed):
Lgtm

devidask27 (approved):
LGTM
Before merging please test it on supported version of ubuntu, centos and rhel

pkg/resmgr/resmgr.go
```diff
@@ -57,6 +57,10 @@ func (c *ResmgrImpl) AuthorizeHost(hostID string, token string) error {
 	client.Logger = &util.ZapWrapper{}
 
 	url := fmt.Sprintf("%s/resmgr/v1/hosts/%s/roles/pf9-kube", c.fqdn, hostID)
+	if len(version) != 0 {
+		// Only resmgr v2 APIs support specifying a specific version of the role to install
+		url = fmt.Sprintf("%s/resmgr/v2/hosts/%s/roles/pf9-kube/?version=%s", c.fqdn, hostID, version)
```
mayuresh108, line 62: Is there an extra / before the ? that needs to be removed?

pkg/resmgr/resmgr.go
```diff
@@ -57,6 +57,10 @@ func (c *ResmgrImpl) AuthorizeHost(hostID string, token string) error {
 	client.Logger = &util.ZapWrapper{}
 
 	url := fmt.Sprintf("%s/resmgr/v1/hosts/%s/roles/pf9-kube", c.fqdn, hostID)
+	if len(version) != 0 {
```
mayuresh108, line 60: Note that the version we will get here will be the version of machinepool.spec.version which is in the format: v1.23.8-pmk.115 will it work with resmgr, with the extra v

pacharya-pf9: @erwinvaneyk @mayuresh108 I don't think we should be handling such things in pf9ctl. IMO that should be handled in the cluster-api-bootstrap-provider-nodelet controllers when it passes kube version in bootscript.

pkg/resmgr/resmgr.go
mayuresh108: Fixed URL as suggested by pushkar

mayuresh108 (approved):

PR #340, opened 2023-02-17
Missed adding RHEL 8.7 to error message

pshanbhag:
RHEL 8.7 is part of supported OS now.

Needed for PMK-5430


sudhakarg (approved):
lgtm

PR #360, opened 2023-11-08
PMK-6194 : added some more dir in clean list (decommission-node)

devidask27:
## ISSUE(S):

[PMK-6194](https://platform9.atlassian.net/browse/PMK-6194)
## SUMMARY

Decommission-node comand was not cleaning all directories. so add some more dir in list which gets cleaned as part of decommission-node. 
## ISSUE TYPE

decommission-node command

#### Reviewers

/cc @psarwate @poojaghumre @mithilarun

### DIFF
--- a/pkg/pmk/decomissionNode.go
+++ b/pkg/pmk/decomissionNode.go
@@ -14,14 +14,8 @@ import (
 )
 
 func removePf9Installation(c client.Client) {
-	fmt.Println("Removing /etc/pf9 logs")
-	cmd := fmt.Sprintf("rm -rf %s", util.EtcDir)
-	c.Executor.RunCommandWait(cmd)
-	fmt.Println("Removing /var/opt/pf9 logs")
-	cmd = fmt.Sprintf("rm -rf %s", util.OptDir)
-	c.Executor.RunCommandWait(cmd)
 	fmt.Println("Removing pf9 HOME dir")
-	cmd = fmt.Sprintf("rm -rf $HOME/pf9")
+	cmd := fmt.Sprintf("rm -rf $HOME/pf9")
 	c.Executor.RunCommandWait(cmd)
 }
 
@@ -54,6 +48,15 @@ func removeHostagent(c client.Client, hostOS string) {
 		cmd := fmt.Sprintf("rm -rf %s", file)
 		c.Executor.RunCommandWait(cmd)
 	}
+	fmt.Println("Running clean all")
+	if hostOS == "debian" {
+		cmd := fmt.Sprintf("apt-get clean all")
+		c.Executor.RunCommandWait(cmd)
+	} else {
+		cmd := fmt.Sprintf("yum clean all")
+		c.Executor.RunCommandWait(cmd)
+	}
+
 }
 
 func DecommissionNode(cfg *objects.Config, nc objects.NodeConfig, removePf9 bool) {
--- a/pkg/util/constants.go
+++ b/pkg/util/constants.go
@@ -124,7 +124,7 @@ func init() {
 	RequiredPorts = []string{"443", "2379", "2380", "8285", "10250", "10255", "4194", "3306", "8158", "5672", "5673", "8023", "9080", "6264", "5395", "8558"}
 	ProcessesList = []string{"kubelet", "kube-proxy", "kube-apiserver", "kube-scheduler", "kube-controller", "etcd"}
 	Pf9Packages = []string{"pf9-hostagent", "pf9-comms", "pf9-kube", "pf9-muster"}
-	Files = []string{"/opt/cni", "/opt/containerd", "/var/lib/containerd", "/var/opt/pf9"}
+	Files = []string{"/opt/pf9", "/etc/pf9", "/var/opt/pf9", "/var/log/pf9", "/tmp/*", "/var/spool/mail/pf9", "/opt/cni", "/etc/cni", "/var/log/messages-*", "/var/lib/docker", "/run/containerd", "/opt/containerd", "/var/lib/containerd", "/var/lib/nerdctl"}
 
 	AzureContributorID = "b24988ac-6180-42a0-ab88-20f7382dd24c"
 


pkg/pmk/decomissionNode.go
```diff
@@ -14,14 +14,8 @@ import (
...
 func removePf9Installation(c client.Client) {
-	fmt.Println("Removing /etc/pf9 logs")
-	cmd := fmt.Sprintf("rm -rf %s", util.EtcDir)
-	c.Executor.RunCommandWait(cmd)
-	fmt.Println("Removing /var/opt/pf9 logs")
-	cmd = fmt.Sprintf("rm -rf %s", util.OptDir)
-	c.Executor.RunCommandWait(cmd)
 	fmt.Println("Removing pf9 HOME dir")
-	cmd = fmt.Sprintf("rm -rf $HOME/pf9")
+	cmd := fmt.Sprintf("rm -rf $HOME/pf9")
```
mithilarun, line 18:
```suggestion
        homeDir, err := os.UserHomeDir()
        if err != nil {
        ....
        }
        pf9Home := path.Join(homeDir, "pf9")
	cmd := fmt.Sprintf("rm -rf %s", pf9Home)
```

mithilarun: Why are the other commands removed?

devidask27: > Why are the other commands removed?

I added this repositories in list here
```go
Files = []string{"/opt/pf9", "/etc/pf9", "/var/opt/pf9", "/var/log/pf9", "/tmp/*", "/var/spool/mail/pf9", "/opt/cni", "/etc/cni", "/var/log/messages-*", "/var/lib/docker", "/run/containerd", "/opt/containerd", "/var/lib/containerd", "/var/lib/nerdctl"}
```

pkg/pmk/decomissionNode.go
```diff
@@ -14,14 +16,13 @@ import (
...
-	cmd := fmt.Sprintf("rm -rf %s", util.EtcDir)
-	c.Executor.RunCommandWait(cmd)
-	fmt.Println("Removing /var/opt/pf9 logs")
-	cmd = fmt.Sprintf("rm -rf %s", util.OptDir)
-	c.Executor.RunCommandWait(cmd)
 	fmt.Println("Removing pf9 HOME dir")
-	cmd = fmt.Sprintf("rm -rf $HOME/pf9")
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		zap.S().Debugf("Failed to get home dir, could not delete pf9Home dir")
```
mithilarun, line 22: Please change this log level to error, and return.

mithilarun (changes requested):

pkg/pmk/decomissionNode.go
devidask27: Done

pkg/util/constants.go
```diff
@@ -124,7 +124,7 @@ func init() {
 	RequiredPorts = []string{"443", "2379", "2380", "8285", "10250", "10255", "4194", "3306", "8158", "5672", "5673", "8023", "9080", "6264", "5395", "8558"}
 	ProcessesList = []string{"kubelet", "kube-proxy", "kube-apiserver", "kube-scheduler", "kube-controller", "etcd"}
 	Pf9Packages = []string{"pf9-hostagent", "pf9-comms", "pf9-kube", "pf9-muster"}
-	Files = []string{"/opt/cni", "/opt/containerd", "/var/lib/containerd", "/var/opt/pf9"}
+	Files = []string{"/opt/pf9", "/etc/pf9", "/var/opt/pf9", "/var/log/pf9", "/tmp/*", "/var/spool/mail/pf9", "/opt/cni", "/etc/cni", "/var/log/messages-*", "/var/lib/docker", "/run/containerd", "/opt/containerd", "/var/lib/containerd", "/var/lib/nerdctl"}
```
psarwate, line 127: Are we deleting /var/log/messages-*? Why has this been added to the files?

psarwate: @mithilarun any idea why /var/log/messages-* need deletion? These are system log files, other node components would have logged here as well. We should not be trouching these IMO.

psarwate: Rest of the changes look fine to me @devidask27

anmolsachan:
@devidask27 @psarwate Does this cover etcd cleanup as well if the node is a master node? 
Please see https://platform9.atlassian.net/browse/DOCS-151. 
IMO we should add a master=true/false(default) flag for the detach/decommission command as well so that it removes the master node as etcd member as well and does cleanup.

devidask27:
> @devidask27 @psarwate Does this cover etcd cleanup as well if the node is a master node? Please see https://platform9.atlassian.net/browse/DOCS-151. IMO we should add a master=true/false(default) flag for the detach/decommission command as well so that it removes the master node as etcd member as well and does cleanup.

@anmolsachan this not covers etcd cleanup. decommision-node not works for nodes which are part of cluster user have to remove it before running decommision-node. so i think no need to check if node is master or not

psarwate:
> @devidask27 @psarwate Does this cover etcd cleanup as well if the node is a master node? Please see https://platform9.atlassian.net/browse/DOCS-151. IMO we should add a master=true/false(default) flag for the detach/decommission command as well so that it removes the master node as etcd member as well and does cleanup.

@anmolsachan the JIRA talks of forced removal of the nodes. In the event node has to be removed forcibly due to hardware failure, we are perhaps just missing steps to run "etcdctl remove member" from a healthy node. Don't think such cleanups should be clubbed with pf9ctl decommission-node.
If required we can look at adding a new cli option to pf9ctl that can specifically do this.

pkg/util/constants.go
mithilarun: @psarwate Copied that list from the directories that the SA teams currently clean up. Let's skip it.

psarwate: @devidask27 lets remove  /var/log/messages-*.

devidask27: Done

psarwate (approved):
/lgtm

poojaghumre (approved):

mithilarun (approved):

PR #364, opened 2023-12-19
Bump attach node retry attempts

pshanbhag:
Previously we would try 6 times which is about 3 mins with the 30 secs sleep between retries. However, this may not be long enough because after a prep-node, the host goes through a kube role install and till that completes an attach won't be successful. If users try to do the attach as soon as prep-node completes (say through automation), the 3 minute may not be sufficient time for the attach to succeed. Changing it to about 9 mins which should be a very conservative estimate.


xagent003 (approved):
/

devidask27 (approved):
/lgtm

PR #407, opened 2025-03-13
[PMK-6629] send projectID during resmgr authorizeHost call

deepak07dbz:
## ISSUE(S):
https://platform9.atlassian.net/browse/PMK-6629
projectID is not getting reflected in qbert DB after prep-node, due to which nodes are visible in all the tenants before cluster creation.

## SUMMARY
send projectID during resmgr authorizedHost call during prep-node

## DEPENDS ON:
https://github.com/platform9/pf9-qbert/pull/2352

#### Manual
shared in above PR


mridulgain (approved):
/lgtm
/approve

mithilarun (approved):

